PR #5, opened 2020-12-17
Updates to the examples and the documentation

olofmansson:
All the examples now use the provided example data and the Rd files in the man folder has be regenerated.


olofmansson (commented):

PR #7, opened 2020-12-28
Update Olink_normalization.R

olofmansson:
- Fixes a problem with flagged and missing bridge samples
 - Removed Panel variable from select() just before spread() to deal with NPX files with conflicting Panel names (Panel names updated in 10/1/2020 release of NPX Manager) 

Co-authored-by: @jbarbagallo


olofmansson (commented):

PR #15, opened 2021-06-29
imported packages selectively and ref in function for pca_plot()

pascal-git:


kathy-nevola:
PCA function passes testthat unit testing and all changes seem to work as intended.

PR #16, opened 2021-07-02
Github workflow to test that `remotes::install_local()` works

AskPascal:
This code will test if the package is installable using `remotes::install_local()` which is very much parallel to the `devtools::install_github()` method we currently recommend in the readme.

Tests are run on Windows / Ubuntu / Mac and with R version 4.0.0 and the most recent release.

The runtime of the workflow is rather long since many packages need to be installed. I implemented caching on Mac and Linux however.

Best time to merge this PR might be after merging the function specific reformatting branches. 

Also: I'm sorry for the messy commit history, it was a bumpy ride to make sure the github engines liked my code... Might be good to squash the commits before merge...

### DIFF
--- a/.github/workflows/github-actions-install.yml
+++ b/.github/workflows/github-actions-install.yml
@@ -3,6 +3,9 @@ on: [push]
 jobs:
   Can-it-be-installed:
     runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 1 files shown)


AskPascal:
Superseeded by #31

PR #19, opened 2021-07-12
Format boxplot to import selectively

kathy-nevola:
Boxplot has been formatted to import selectively. However there is no unit test, so I tested with the example code against current version of OlinkAnalyze.


marisand (approved):
It looks good and works as expected.

PR #20, opened 2021-07-12
Format bridgeselector to import selectively

kathy-nevola:
Bridge selector function has been formatted to import selectively. Unit test passed.


marisand (approved):
Added some small changes only.

PR #25, opened 2021-07-30
Test that a PCA plot with example data looks as expected

AskPascal:
Using `testthat` and `vdiffr` to ensure that a PCA plot based on example data is identical to the expectation.

to run the tests independent from `devtools::check()` use `devtools::test()`


marisand:
Closing after agreement to put with other testfiles

PR #30, opened 2021-08-06
Feature: Use Swedish Gothic Thin for plot labels if available

AskPascal:
Modified the `set_plot_theme()` function to accept a `font` parameter (default: "Swedish Gothic Thin") that is used for any text in the plots.

If {extrafont} or the requested font family are not available the ggplot default font is used without any error message.


marisand:
Works as expected for me. Merging.

PR #42, opened 2021-08-30
Panel name set to Olink PANEL NAME, remove T96

kathy-nevola:
Panel naming should now be Olink PANEL NAME. Removed any occurrence of Target 96 and set the last word to be all capitalized. Tagging @marisand since you have the most knowledge of the panel names (thanks for getting that information together) and @simfor as these may impact unit tests.


kathy-nevola:
Issue: Due to the renaming of the T96 panels, projects that are bridged between old and newly named panels will be treated separately in the olink_ttest since it groups by Panel. The OlinkIDs are the same but panel names are different.  This should fix changes in T96 panel names.

marisand:
LGTM

PR #44, opened 2021-08-31
converted to title case, except Roman Numerals

kathy-nevola:
Changed read NPX function to convert from all capitals to Title case.

For example:
Olink Target 96 Inflammation -> Olink Inflammation
Olink Target 96 Cardiovascular II -> Olink Cardiovascular II
Olink INFLAMMATION -> Olink Inflammation
Olink CARDIOVASCULAR II -> Olink Cardiovascular II


marisand:
looks fine

PR #47, opened 2021-09-01
PlateSize for ranomizer

marisand:
PlateSize variable added to take plate sixes for T48 and T96. Easily generelizable to other plate sizes and hardcoded sizes removed from code.


Orbmac (commented):
Looks good. But we probably should remove the Groups argument in the function and in the help function.

Orbmac:
Looks good now without the group factor.

PR #48, opened 2021-09-01
global.r: fix documentation

AskPascal:
This will prevent roxygen from trying to document the content in that file as a function

Also rerun roxygen document and updated Rproj file to include a documentation run on every devtools::check() to flag issues related to roxygen earlier


AskPascal:
local `devtools::check()` passed

OlinkAnalyze/R/globals.R
```diff
@@ -1,13 +1,12 @@
-#' Global Variables Defined
-#' @importFrom utils globalVariables
```
AskPascal, line 2: Since we use globalVariables only with the namespace added (`utils::globalVariables`) this import is not required.

kathy-nevola (approved):
LGTM thanks!

Issue #17, opened 2021-07-12
Plotting functions with scales causes warning


kathy-nevola:
Updated version of ggplot2 changes how scale is used (ggplot2_3.3.5)
> olink_pca_plot(npx_data1 %>% filter(!str_detect(SampleID,"CONTROL")))
Warning message:
`guides(<scale> = FALSE)` is deprecated. Please use `guides(<scale> = "none")` instead. 

> print(sessionInfo())
R version 4.0.5 (2021-03-31)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19041)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                           LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] OlinkAnalyze_1.2.4 scales_1.1.1       RColorBrewer_1.1-2 readxl_1.3.1       forcats_0.5.1      dplyr_1.0.7       
 [7] purrr_0.3.4        readr_1.4.0        tidyr_1.1.3        tibble_3.1.2       tidyverse_1.3.1    stringr_1.4.0     
[13] openxlsx_4.2.4     lmerTest_3.1-3     lme4_1.1-27.1      Matrix_1.3-2       ggrepel_0.9.1      ggfortify_0.4.12  
[19] ggplot2_3.3.5      emmeans_1.6.2-1    car_3.0-11         carData_3.0-4      broom_0.7.8       

loaded via a namespace (and not attached):
 [1] nlme_3.1-152        fs_1.5.0            usethis_2.0.1       lubridate_1.7.10    ordinal_2019.12-10  devtools_2.4.2     
 [7] httr_1.4.2          rprojroot_2.0.2     numDeriv_2016.8-1.1 tools_4.0.5         backports_1.2.1     utf8_1.2.1         
[13] R6_2.5.0            DBI_1.1.1           colorspace_2.0-2    withr_2.4.2         tidyselect_1.1.1    gridExtra_2.3      
[19] prettyunits_1.1.1   processx_3.5.2      curl_4.3.2          compiler_4.0.5      extrafontdb_1.0     rvest_1.0.0        
[...]

PR #59, opened 2021-09-23
Bugfix develop

kathy-nevola:
fix bug in pca plot so the correct version is used in develop/production


AskPascal (approved):
Looks good to me, just re-added the "guides" fix from issue #17 ...

marisand:
Agree with AskPascal.

PR #60, opened 2021-09-24
Randomized order output bridge selector

marisand:
Removing Order from output and randomizing output so bridge samples are not in decreasing mean NPX order.


AskPascal (approved):
Looks good to merge, but CI failing due to unrelated problems on develop. Therefore we should fix them first...

PR #76, opened 2021-11-18
Decrease size of data in package

kathy-nevola:
edited npx_data1 and npx_data2 to decrease size. Made changes to unit tests, vignette, and help documentation to reflect this change


AskPascal (approved):
Looks good to merge. 

Just not 100% sure about the effects of changing filtration from `filter(Threshold == 'Significant')` to `head(10)` in some tests, but this doesn't change the code behavior.

simfor (approved):
Looks great! We could probably also remove /inst/extdata/npx_data2... entirely

PR #83, opened 2021-12-09
Fixes #79

### LINKED ISSUE #79: Handle NPX data sets with additional metadata columns
Parsing fails when trying to load NPX data with additional columns at the end of the data file.  If these columns are manually removed, the parser works fine.

```
dat <- OlinkAnalyze::read_NPX(filename = "input.xlsx")
```

The error:
```
> rlang::last_error()
<error/tibble_error_na_column_index>
Can't use NA as column index with `[` at positions 1, 2, 3, and 4.
Backtrace:
 1. OlinkAnalyze::read_NPX(filename = "input.xlsx")
 3. tibble:::`[.tbl_df`(meta_dat, 2, control_index)
```

schaffman5:
Modified Read_NPX function to accommodate NPX Excel data files with extra sample metadata (fixes #79).  This additional sample metadata, if present, will be appended to the long formatted output data.  This has been tested on 4 NPX Excel files with and without these additional data columns.


kathy-nevola:
Meta data should be imported separate from NPX and merged within R as opposed to changing NPX file externally for reproducibility and trackability.

PR #85, opened 2021-12-21
Bugfix: Trademark changes

kathy-nevola:
In all mentions of the package we write it as Olink:tm: Analyze in Readme and Vignette
Remove the hexagon sticker
This change will merge into both Main and Develop


AskPascal:
Setting this to draft for now since we agreed to merge this only to develop, right?

PR #86, opened 2021-12-21
Bugfix: Trademark changes

kathy-nevola:
In all mentions of the package we write it as Olink® Analyze in Readme and Vignette
Remove the hexagon sticker
This change will merge into both Main and Develop

Tied to PR #85


AskPascal (commented):
Tiny suggestions only

AskPascal (approved):

PR #96, opened 2022-01-21
Use TRUE/FALSE instead of T/F

AskPascal:
Improves consistency and prevents problems if T or F are used as variable names.


kathy-nevola (approved):
LGTM

PR #98, opened 2022-01-21
Plate randomizer doc updates

simfor:


OlinkAnalyze/R/Olink_plate_randomizer.R
```diff
@@ -151,20 +151,42 @@ generatePlateHolder <- function(n.plates,n.spots,n.samples, PlateSize){
...
+#'
+#' The generated plating scheme can be visualized using the function \code{\link{olink_displayPlateLayout}}.
+#' How well the relevant variables are randomized can be visualized using the function \code{\link{olink_displayPlateDistributions}}.
 #' @param Manifest tibble/data frame in long format containing all sample ID's. Sample ID column must be named SampleID.
 #' @param PlateSize Integer. Either 96 or 48. 96 is default.
 #' @param SubjectColumn (Optional) Column name of the subject ID column. Cannot contain missings. If provided, subjects are kept on the same plate.
 #' @param iterations Number of iterations for fitting subjects on the same plate.
-#' @param available.spots Integer. Number of wells available on each plate. Maximum 40 for T48 and 88 for T96. Can also take a vector equal to the number of plates to be used indicating the number of wells available on each plate.
+#' @param available.spots Integer. Number of wells available on each plate. Maximum 40 for T48 and 88 for T96. Takes a vector equal to the number of plates to be used indicating the number of wells available on each plate.
```
kathy-nevola, line 164: Change Integer to Numeric? Not sure if the class changes with a vector or not.

kathy-nevola (changes requested):

OlinkAnalyze/R/Olink_plate_randomizer.R
simfor: Either numeric or integer works here (does it ever make a difference in R? I don't think I ever distinguish between the two :) ). Since numeric is the most common type, I agree that we should probably write that instead

simfor: Maybe it would also be good to add an example on how to use this argument?

kathy-nevola (approved):
Look good to me! Thanks for making those changes!

PR #110, opened 2022-02-16
Checking one datapoint in level for ttest as well as all NA:s

marisand:


kathy-nevola:
Looks pretty good to me, just 1 question. 
Tested:
`data1<- npx_data1 %>% filter(Assay == "CHL1") %>% filter(!(Treatment == "Treated" & SampleID != "A16"))
data2 <- npx_data1 %>% filter(Assay != "CHL1")
data3 <- rbind(data1, data2)
olink_ttest(data3, variable = "Treatment")`

which returned results with warning message: 
`Warning message:
The assays OID01216 have too few datapoints in one level of the factor. They will not be tested. `

However: If just 1 protein is run Warning message will not be output, rather will return the following:
`olink_ttest(data1, variable = "Treatment")
Variable converted from character to factor: Treatment
T-test is performed on Treated - Untreated.
Error in t.test.formula(NPX ~ Treatment, data = ., ...) : 
  grouping factor must have exactly 2 levels
In addition: Warning message:
 Error in t.test.formula(NPX ~ Treatment, data = ., ...) : 
grouping factor must have exactly 2 levels`

Is this the expected or desired behavior?

kathy-nevola (commented):
One question but otherwise looks good, see comment above

kathy-nevola (commented):
Looks good to me! Thanks for the update.

PR #115, opened 2022-03-18
Read compressed npx

### LINKED ISSUE #80: Handle gzipped Explore long format csv files
The parser handles the long format of the csv data files from the Explore platform.  However, because of data redundancy and resulting large file sizes, it would be nice if the parser would handle gzipped versions of these csv files directly.

klevdiamanti:
Allowing *.zip files exported from MyData to be read in R.
Fixes #80


MasoumehSheikhi (changes requested):
I looked at a few zip files from mydaya and there are some with checksum file named different from MD5_checksum.txt (I found a few with checksum_sha256.txt). That makes [line 64] length(compressed_file_csv)!=1.

MasoumehSheikhi (approved):
We might want to have an eye for possibly different checksum file names in future.

PR #127, opened 2022-04-20
Added heatmap function

Orbmac:
Added a heatmap function which uses pheatmap.
Added pheatmap to DESCRIPTION.


MasoumehSheikhi (commented):
In heatmap function:
[70] dplyr::n()
[115] dplyr::filter
[119] dplyr::mutate
[143] dplyr::select
[145] tibble::column_to_rownames

PR #135, opened 2022-05-12
Change namespace for unzip

klevdiamanti:
Using zip::unzip instead of utils::unzip as this was failing in RStudio Workbench


AskPascal:
You will need to add the `zip` package to DESCRIPTION too: https://r-pkgs.org/description.html?q=depende#description-dependencies

klevdiamanti:
Fixing it asap.

PR #136, opened 2022-05-17
Use external action for commit and push in `document` workflow

AskPascal:


AskPascal:
Was not working as expected

PR #140, opened 2022-05-24
Do not run render-and-document workflow on main

AskPascal:
since main is protected, this workflow is expected to fail anyway...

example of this happening here: https://github.com/Olink-Proteomics/OlinkRPackage/runs/5839141095?check_suite_focus=true


kathy-nevola (approved):
LGTM

Issue #112, opened 2022-02-17
Wrong output for olink_dist_plot() in vignette


OlaCaster:
The vignette states that `olink_dist_plot()` outputs a list of `ggplot` objects, when in fact it directly returns a `ggplot`. The function documentation is correct.

PR #141, opened 2022-05-25
Notes 3.1

kathy-nevola:
Fixing notes for devtools::check and updating news.

Also fix Issue #112


kathy-nevola:
@klevdiamanti I noticed the ordering on the figures for pathway enrichment wasnt working so I fixed it. Can you please review?

klevdiamanti:
> @klevdiamanti The 3K.zip file was too large so I had to remove it from the build and comment out the associated tests. Let me know if you would like to replace it with a smaller datefile to test. From the devtools::test_coverage, this test only added 0.5% more coverage to the unit tests (from 70.79% to 70.41%)

I think we can release as is, and I will make a new smaller test file. Apologies for this, I did not think about file size when implementing.

klevdiamanti:
> @klevdiamanti I noticed the ordering on the figures for pathway enrichment wasnt working so I fixed it. Can you please review?

These modifications look great to me and they work like a charm. I am not merging as others mentioned that they might take a look at this PR.

PR #156, opened 2022-06-17
NPX read to handle Panel_Version or Panel_Lot_Nr

klevdiamanti:
# Title: read_NPX to handle colnames not matching old or new colnames
**Problem:** read_NPX was crashing when Panel_Version was named Panel_Lot_Nr while colanmes were still the same as the old expected colnames. This was not exactly a bug as we were not aware of the existance of this combination, but we are fixing it so that it can be handled as a bug.

I used this chance to shorten the example datasets used for unit testing, and un-commented the tests for it.

**Solution:** Changed it so that read_NPX can accept either Panel_Version or Panel_Lot_Nr.

## Checklist

## Type of changes

- [x] Bugfix (non-breaking change which fixes an issue) (link the issue on the right)
- [ ] Documentation Update 
- [x] Other (allow unit test for large test file to work again)


AskPascal (changes requested):
There is also an updated needed here imho:
https://github.com/Olink-Proteomics/OlinkRPackage/blob/9d7fdb641cf26461f3843dadfe342ecd7ac5038b/OlinkAnalyze/R/Read_NPX_data.R#L146-L153

AskPascal (approved):
LGTM

PR #170, opened 2022-08-16
Make read_NPX() warn on projects were certain samples were not run for a certain panel

AskPascal:
**Problem:** Copy from Kanban board: _In cases where some samples are run on some, but not all, panels on Target 96 samples will be read in with NA in NPX and will look like samples ran on all panels. May be useful for read_NPX() to throw a warning when something has NA in the NPX column so we know we need to look into it more._
_"The following samples have NA NPX values for the following panels" or something like that?_

**Solution:** Implemented some code to check for these cases and throw a clear warning to inform the user that further checks might be required

## Checklist
- [ ] Make sure you are make a pull request against either **main or develop** (left side). (Requesting to main should be reserved for bugfixes and new releases)
  - Im suggesting to merge this into another branch that contains changes to the same functions

## Type of changes

- [ ] Documentation Update 
- [ ] Other (explain)


MasoumehSheikhi (commented):
There are a bunch of functions with changes. I marked the ones that I reviewed but I'm not sure if it would show you exactly which one was viewed by me. Those that I reviewed looked good and worked with my examples. Please let me know if it didn't tell you which ones I had marked as viewed.

PR #206, opened 2022-12-13
PCA index

amrita-kar:
# Title: PCA plot removing the Index dependence 
**Problem:** The PCA plot fails when there is a interchange between Index and SampleID.

**Solution:** I changed the npxProcessing.R function to group_by only SampleID and not Index field anymore. To check this I randomized the Index field to contain SampleIDs instead of Index. As a result, even if there is an interchange between SampleID and Index the npx processing can happen with SampleIDs only.

**Key Features:**

* The above changes were tested on the PCA plot only for now comparing the old PCA minimal PCA plot with the PCA plot with index dependence removed 

## Checklist

## Type of changes

- [ ] Documentation Update 
- [ ] Other (explain)

## Further comments

- I have only tested the PCA plots on unit testing. This may affect other plots too
- I will be testing this fix also on other plots like UMap and some other plots maybe (need to check on which ones relate to Index issue).


OlinkAnalyze/tests/testthat/test-pca_plot.R
```diff
@@ -218,6 +224,18 @@ test_that("minimal PCA plot", {
 
   vdiffr::expect_doppelganger("PCA plot - label outliers", pca_plot_outliers[[1]])
 
+  #Removing Index dependence in PCA plot
+  ids <- sample(npx_data1$SampleID, 100, replace = T)
+  ids <- gsub("A|B","",ids)
+  ids <- unique(sort(as.numeric(ids)))
```
AskPascal, line 230: There is a control sample in the list of ids at this stage which is not numeric but is called "CONTROL_SMPLE_S" which leads to a warning here...

MasoumehSheikhi (dismissed):
- [x] In "npxProcessing.R" file, remove Index from lines 56 and 160

- [x] Example: npx_data1 |> filter(!str_detect(SampleID,'CONT')) |> mutate(Index=if_else(Panel=='Olink Cardiometabolic', Index+1L, Index)) |>
     olink_pca_plot(quiet = TRUE)

amrita-kar:
> * [x]  In "npxProcessing.R" file, remove Index from lines 56 and 160
> * [ ]  Example: npx_data1 |> filter(!str_detect(SampleID,'CONT')) |> mutate(Index=if_else(Panel=='Olink Cardiometabolic', Index+1L, Index)) |>
>   olink_pca_plot(quiet = TRUE)

amrita-kar:
Also deleted 'Index' on line 157

AskPascal:
As we now rely on SampleID being unique. Should we check for that somewhere? Or do we do that already?

Something along these lines might be helpful:
```
df %>% group_by(SampleID, OlinkID) %>% tally()
```

PR #220, opened 2023-01-26
allow read_NPX support extended and simple NPX files

klevdiamanti:
# Title: Allow read_NPX to read extended NPX files
**Problem:** After the latest update of OA read_NPX was crashing when reading in extended NPX files. This was because the extended NPX files contained a column called Sample_Type, but no column called ExploreVersion. read_NPX expected an exact match that was not there.

**Solution:** Extnded NPX column names were added to the function. Also, read_NPX does not crash when unexpetded column names are provided.

**Key Features:**

* read_NPX does not crash when unexpetded column names are provided
* read_NPX natively handels EXTENDED NPX files

## Checklist

## Type of changes

- [X] Code style update (formatting, renaming)
- [X] Documentation Update 
- [ ] Other (explain)


Orbmac (commented):
Seem to be working fine. I first saved a test extended file as XLSX but noticed that wont work. But maybe we cant even save explore as ELSX files.

PR #294, opened 2023-11-02
News and Description Update

kathy-nevola:
Minor update to news and description in preparation for next release


amrita-kar (approved):
Went through the OA board to check the changes.

PR #388, opened 2024-06-20
LOD updates

kathy-nevola:
Functionality added to improve usability and fix minor bugs including:

- Updated column names to match FixedLOD file
- Removal of Assay Warning == Pass as a criteria
- Addition of PlateID as a grouping variable in pc median calculation

### DIFF
--- a/OlinkAnalyze/NEWS.md
+++ b/OlinkAnalyze/NEWS.md
@@ -1,7 +1,7 @@
 # Olink Analyze 3.8.0
 ## Minor Changes
 * Function support for calculating LOD for Explore data (#355, @kathy-nevola, @klevdiamanti, @leiliuC)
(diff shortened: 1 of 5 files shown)


amrita-kar:
Checked the olink_lod.R code for FixedLOD with an old dataset. Had some issues...but I might be doing something wrong. Will check-in in person to discuss.  

- Typo on line 68 
"# If NPX is intensity normalized, than intensity normalize LOD" -- > # If NPX is intensity normalized, then intensity normalize LOD
- On the Vignette for LOD it was a little unclear why the table below is labelled PCNormalizedLOD instead of IntensityNormalizedLOD where the text mentions that for an intensity normalized dataset the LOD column will include LOD values based on the intensity normalized NPX values. 
image

amrita-kar:
image
For fixedLOD, lod_count assays have NA in LOD column

amrita-kar (approved):

PR #422, opened 2024-08-03
3k-HT mapping file

klevdiamanti:
# Title: Add 3k-HT mapping file
**Problem:** The new feature to bridge 3k-HT will require a dataset indicating the mapping of assays between the two platforms.

**Solution:** We have created the file internally and it is added to the feature branch.

**Key Features:**

* Allows mapping of assays between E3072 and EHT

## Checklist
- [ ] Add or update unit tests (if applicable)

## Type of changes

- [ ] Documentation Update 
- [ ] Other (explain)

## Further comments
None


klevdiamanti:
Please check my comments in our Teams chat "Olink Analyze Weekly Meeting" and the suggestions in the code that generates the datasets, and consider updating the "OlinkIDMapping.rds" file.

OlinkAnalyze/inst/extdata/OlinkIDMapping.rds
klevdiamanti, line 1: Data stored under "inst/extdata" are meant to be example data that external users can use. I am thinking that it might be better to store this dataset under "data-raw" as this is meant to host scripts and data that the package should use internally, and they are not meant to be used by external users. This is my interpretation of building R packages, and it is not necessary to implement now. If we decide to go this way later, it will be very easy to change it.

MasoumehSheikhi (approved):

OlinkAnalyze/data-raw/mapping.R
```diff
@@ -0,0 +1,10 @@
+# code to prepare `mapping` dataset goes here
+# Raw data is generated here:
+# Data_Science/Internal Projects/3k_to_HT_bridging/R code/Bridgeable/mapping_file.R
```
kathy-nevola, line 3:
```suggestion

```

PR #431, opened 2024-08-20
change to fread to support older files

kathy-nevola:
it seems that with version 3.9.0, no CSV files with quotation marks in them can be imported anymore. there are still lots of old files that have them. It would also seem difficult to import files then where for example SampleID contains comma or semi-colon.

the underlying problem is that utils::read.table is simply very inflexible and error-prone when the input files can have a mixture of formatting. In my experience, data.table::fread(file="<..>", data.table=FALSE) works a lot better with its auto-detection methods, meaning you do not need to specify what 'sep' and 'quote' should be


OlaCaster (approved):
I played around a little bit with the `data.table::fread` arguments, and I think what I ended up with emulates the old import very well (see below). The major upsides are:

- We don't need to specify `quote` (`fread` will automatically try the data both as quoted and as unquoted)
- We don't need to specify `sep` since it is auto-determined by `fread`
- This is much more performant than `read.table` especially on larger CSV files (see below)

There is a downside to the change, which is the need to add a new dependency. However, `data.table` is a common package in the ecosystem, meaning that a lot of users would already have it anyway.

I performed an automated testing of seven files across versions 3.8.2, 3.9.0 and this `data.table`-version. Only one of these files was formatted according to current export software, the others were legacy files of varying kinds. The overall results were the following:

- v 3.9.0 was able to read 4 of 7 files
- v 3.8.2 was able to read 5 of 7 files
- This `data.table` version was able to read 6 of 7 files
- The one file that no version read was a wide format Signature CSV. This is not formally supported, but does exist (rarely) in the wild.
- The one file that only the `data.table` version could read was a T48 quant long format file (also some kind of legacy format).
- In all cases where I could compare (i.e., when both DT and the existing version imported the data), the resulting objects were identical. (Meaning they gave `TRUE` back for `identical(x,y)` checks.) The caveat to this is that for two of the files I had to replace `X` by `V1` (in version 3.8.2) as assigned column names when it was missing in the CSV header. This difference is inconsequential in my opinion.

In terms of benchmarking, I compared importing two large files (HT with 100 or 500 samples) between version 3.8.2 and this one. Looking at median times over five repetitions (on my computer) the 100 sample file went from 8.1s to 1.8s while the 500 sample file went from 45.3s to 7.7s.

Overall, considering this version could import more files successfully than either 3.8.2 or 3.9.0 plus the performance advantages, I think this is a very worthwhile patch that will also save us unnecessary errors in Insight downstream.

I have one question mark (see separate comment below).

OlaCaster:
I am uncertain about this test which is there from before:

```r
  # if this fails too, then throw an error
  if (is.data.frame(df_npx) &&
      ncol(df_npx) == 1) {

    stop("Could not read NPX csv file! Delimiter is not \";\" or \",\". ")

  }
```

I don't think `fread` would return a one-column data set, but I am not sure. (I think it would rather error out if the delimiter could not be determined for example.) Anyway, I would suggest removing the second sentence here about the delimiter, since this new import would work for any kind of delimiter used in the file.

kathy-nevola:
@OlaCaster , this PR adds an additional dependency to Olink Analyze. We are limited in the number of dependencies available unfortunately.

klevdiamanti (approved):
LGTM

PR #439, opened 2024-08-30
Develop 4.0 QS normalization

amrita-kar:
**Key Features:**

* Added QS normalization function (olink_normalization_quantile_smoothing.R)
* Added tests/testthat/test-olink_normalization_quantile_smoothing.R with basic unit tests 

## Checklist

## Type of changes

- [ ] Documentation Update 
- [ ] Other (explain)

## Note
This branch might also have the updates from bridgeable function so need to fix that


amrita-kar:
- Fixed some linting in qs normalization code. I also pushed those changes to the branch.  
- Only line 73 and 97 have some linting issues now. 
- I was checking the inputs to the qs function and it mentions that it needs the ref_cols. Do they need to be named the same for the function to work. For example if I change $uniprot to [1] "Gene" I still get a qs transformed matrix. Does that matter?

klevdiamanti:
> * Fixed some linting in qs normalization code. I also pushed those changes to the branch.
> 
>     * Only line 73 and 97 have some linting issues now.
> 
>     * I was checking the inputs to the qs function and it mentions that it needs the ref_cols. Do they need to be named the same for the function to work. For example if I change $uniprot to [1] "Gene" I still get a qs transformed matrix. Does that matter?

Point 2 fixed.

The function "olink_normalization" function has a function called "norm_internal_rename_cols" where it changes the column names of the non-reference dataset to match the ones from the reference. So "olink_normalization_qs" should expect the column names of the two input datasets to match.

OlinkAnalyze/R/olink_normalization_quantile_smoothing.R
```diff
@@ -0,0 +1,245 @@
...
+      Project = names(lst_df)[2L]
+    )
+
+  # ref columns for output ----
+
+  df_ref_output <- lst_df[[1L]] |>
+    dplyr::filter(
+      # only customer samples
+      .data[["SampleType"]] == "SAMPLE"
+      # remove interal control assays
```
MasoumehSheikhi, line 227:
```suggestion
      # remove internal control assays
```

kathy-nevola (commented):
Recommend adding method information: 

`In the case when a study is separated into multiple projects, an additional normalization step is needed to allow the data to be comparable across projects. Across different Olink products, some of the assays exist in corresponding but distinct NPX spaces. For those assays, the median of paired differences is insufficient for bridging as it only considers one anchor point (the median/50% quantile). Instead, quantile smoothing (QS) using multiple anchor points (5%, 10%, 25%, 50%, 75%, 90% and 95% quantiles) is favored to map the Explore 3072 data to the Explore HT distribution. The \`olink_normalization_qs()\` performs quantile smoothing bridging normalization between datasets from two Olink products (for example Olink Explore 3072 and Olink Explore HT) by performing the following steps: \cr

\itemize{

 \item An empirical cumulative distribution function is used to map datapoints for the bridging samples from one product to the equivalent space in the other product.
 \item A spline regression model is constructed using unmapped and mapped data from one product, using anchor points from the quantiles defined above.
 \item The spline regression model is used to predict the normalized NPX values for all datapoints

}

More information on quantile smoothing and between product normalization can be found in the Bridging Olink^®^ Explore 3072 to Olink^®^ Explore HT tutorial.
`

kathy-nevola:
Examples are not technically needed as it is an internal function but it is fine to have anyway. I cant comment on the example as I am not sure how the function works. Note dontrun and donttest will still run on CRAN (and locally I think). Otherwise it looks good to me.

OlinkAnalyze/R/olink_normalization_quantile_smoothing.R
```diff
@@ -0,0 +1,308 @@
...
+#'
+#' @return A "tibble" of Olink data in long format containing both input
+#' datatsets with the quantile normalized quantifications.
+#'
+#' @keywords Normalization; Quantile ; Smoothing
+#'
+#'
+#' @examples
+#' \dontrun{
+#' \donttest{\
```
amrita-kar, line 61:
```suggestion
#' \donttest{\
```
```suggestion
#' \donttest{
```

PR #451, opened 2024-09-17
update faq links with queries

kathy-nevola:
It is now possible to link to an FAQ page with a specific query. FAQ links have been updated.


dtopouza (approved):
All links have been addressed and are functional.

PR #456, opened 2024-09-20
News 4.0

kathy-nevola:
Add news of upcoming release (awaiting final content to confirm CRAN comments)


OlinkAnalyze/NEWS.md
```diff
@@ -1,3 +1,19 @@
+# Olink Analyze 4.0.0
+## Major Changes
+* olink_normalization can now be used for bridging Explore 3072 data to Explore HT data (#453, #452, #439, #449, #447,#446, #441, #440, #434, #422, #436, #435 @klevdiamanti, @kathy-nevola, @kristynchin-olink, @dtopouza, @amrita-kar)
```
dtopouza, line 3: We should add Masoumeh here, maybe in some other points too if she worked on optimization and the tutorial.

OlinkAnalyze/inst/WORDLIST
```diff
@@ -192,4 +207,4 @@ PlateLQL
 Quant
 Topouza's
```
dtopouza, line 208: Is this something we should fix? I don't see anyone else's name, not sure where this came from.

dtopouza (changes requested):

OlinkAnalyze/inst/WORDLIST
kathy-nevola: Its because your name is in the  News (since we added you to the author list)

PR #462, opened 2024-09-24
remove large test files from build

kathy-nevola:
CRAN reports tarball at 7 mb, due to test data being included in submission. As a short term fix, I have skipped tests if the file doesnt exist and added it to .Rbuildignore. Tests passing locally and tarball locally compiles at 4 MB (under 5 MB CRAN limit).

More permanent solution will need to be further investigated for future release.


klevdiamanti (approved):
I pushed a few small commits for tests. Other than that LGTM.

PR #474, opened 2024-11-14
convert block to char when read from CSV

kathy-nevola:
Problem: Currently if Block is 1,2,3... and file is a csv, the input will be int. For parquet and in specification, Block should be char.

Solution: When Block exists, set to character in read_NPX_csv.R

Recreated Previous PR to run correct checks


dtopouza (approved):
Looks good!

PR #479, opened 2024-11-20
Move test data to within testthat file

kathy-nevola:
CI was failing due to not being able to find paths for skip if file doesnt exist tests. Test data recommended to be within testthat folder (ref: https://r-pkgs.org/testing-design.html#storing-test-data) 

Data moved to within testthat folder and test_path added to all paths. All paths updated.


kathy-nevola:
All tests and checks passed locally.

Issue #504, opened 2025-02-03
[BUG] Update MaxLOD after normalization


klevdiamanti:
**Describe the bug**
MaxLOD represents the maximum value of PlateLOD of an assay across plates in a multi-plate project. When we normalize two datasets, MaxLOD is adjusted within each dataset that is normalize, but not across datasets.

**To Reproduce**
Normalize (bridge, subset, median) any 2 multi-plate Target datasets that contain PlateLOD and MaxLOD.

**Expected behavior**
MaxLOD should be updated as the max(MaxLOD) or max(PlateLOD) of each assay.

**System Information:**
 - OS: Any
 - R Version: Any

**Additional context**
Feature was requested by customer.

PR #505, opened 2025-02-03
Update MaxLOD after normalization

klevdiamanti:
# Title: Update MaxLOD after normalization
**Problem:** After normalization MaxLOD was not updated as the MaxLOD across all input datasets.

**Solution:** Implement a function the updates MaxLOD to reflect the correct results of the normalization.

## Checklist

## Type of changes

- [X] Documentation Update 
- [ ] Other (explain)

## Further comments
Addresses #504.


MasoumehSheikhi (commented):
Looks good to me. I just noticed something that is probably not a big deal.

in test-olink_normalization.R line [100]  ### bridge normalization - multiple lod columns 
when I run the test inline, I get an error: Error: `expect_message(...)` did not throw the expected message.

but with Testing R file using 'testthat', all the tests pass.

if this is not problem, then I think it can be merged.

PR #531, opened 2025-03-10
Error if tests have warnings or are skipped

klevdiamanti:
# Title: Error is tests have warnings or are skipped
**Problem:** Given the large size of test and example data in the package, we are often running into trouble to meet the 5MB package size on CRAN. This is mainly caused by data under tests/testtthat/data used for testing and inst/extdata used as example data.

**Solution:** We have removed all test data and almost all example data in the package build to reduce package size. However, because in this case many tests will be skipped due to missing files, we added a workflow that runs all the tests and breaks if any tests are skipped, throw a warning or fail. This ensures that all tests work as expected even in cases when the developer missed to run devtools::test() locally prior to pushing.

**Key Features:**

* New CI worflow.
* Smaller package build size.

## Checklist

## Type of changes

- [ ] Documentation Update 
- [ ] Other (explain)

### DIFF
--- a/.github/workflows/R-CI-test-all.yaml
+++ b/.github/workflows/R-CI-test-all.yaml
@@ -30,21 +30,20 @@ jobs:
           testthat::test_local(path = "OlinkAnalyze", reporter = reporter)
           reporter$end_reporter()
 
-          n_fail <- length(reporter$failures$as_list())
           n_warn <- length(reporter$warnings$as_list())
           n_skip <- length(reporter$skips$as_list())
 
-          if (n_fail > 0L || n_warn > 0L || n_skip > 0L) {
+          if (n_warn > 0L || n_skip > 0L) {
             cli::cli_abort(
               c(
-                "x" = "Identified {.val {n_fail}} failed, {.val {n_warn}}
-                warning, and {.val {n_skip}} skipped tests!",
+                "x" = "Identified {.val {n_warn}} warning(s) and {.val {n_skip}}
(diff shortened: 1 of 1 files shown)


klevdiamanti:
@kathy-nevola is this conceptually similar to #480? If yes, lets discuss which one is to be kept and which to be discarded/

OlinkAnalyze/tests/testthat/test-read_npx.R
kathy-nevola, line 1: is the plan to move these other file types (zip, xlsx, csv) to an external repo for testing?

.github/workflows/R-CI-test-all.yaml
kathy-nevola, line 1: Failures will automatically cause a failure and will not trigger cli message. 
image

kathy-nevola: Skipped tests work as expected. 
image

kathy-nevola: Warnings also work 
image

kathy-nevola (commented):
Overall the check works as expected although any errors will not give the expected message. Might be worth looking into but I am happy to approve it as is.

.github/workflows/R-CI-test-all.yaml
klevdiamanti: You are right about test failures. Great catch. I will the PR to catch only warnings and skips.

PR #540, opened 2025-03-14
Prevent missing parquet meta-data from generating errors

OlaCaster:
# Title: Prevent missing parquet meta-data from generating errors
**Problem:** The previous change in https://github.com/Olink-Proteomics/OlinkRPackage/commit/9c104357a6e2bcad2746b280778ccb3f556ce629 from throwing an error for missing parquet meta-data to just generating a warning does not work as intended. Because of how the subsequent code is written, R itself will generate errors if either `Product` or `DataFileType` are missing (in addition to our own warnings).

If `Product` is missing:
```
Error in if (exp_p_meta$prod %in% p_file_meta & !(meta_[[exp_p_meta$prod]] %in%  : 
  argument is of length zero
In addition: Warning message:
Missing required field "Product" in parquet file's metadata. 
```

If `DataFileType` is missing:
```
ℹ This parquet file is for research use only:
"For Research Use Only. Not for use in diagnostic procedures."!
Error in if (exp_p_meta$dtftp %in% p_file_meta & !(meta_[[exp_p_meta$dtftp]] %in%  : 
  argument is of length zero
In addition: Warning message:
Missing required field "DataFileType" in parquet file's metadata. 
```

**Solution:** The problem with the existing code is that subsequent checks involving `Product` and `DataFileType` include the vectorized boolean operator `&`. This forces R to evaluate the second arguments (`!(meta_[[exp_p_meta$prod]] %in% olink_platforms)` for `Product` and `!(meta_[[exp_p_meta$dtftp]] %in% olink_files)` for `DataFileType`), which results in logical vectors of length 0 when `exp_p_meta$prod` or `exp_p_meta$dtftp` do not exist. This is easy to fix by using the scalar operator `&&` instead, in which case R will skip evaluating the second argument if the first one is `FALSE` (which is the case here). Note that the new code also works even if there are no meta-data at all, so that `exp_p_meta` is `NULL`.

It is generally best practice to use `&&` and `||` instead of `&` and `|` in situations like these, when all arguments are known beforehand to be scalars. It reduces the risk of errors like this and is also slightly more performant since the R interpreter needs to do less work.

**Key Features:**

* No new features (this is a bug fix). The bug has immediate downstream impacts on NPX file import in Insight.

## Checklist
- [ ] Add or update unit tests (if applicable)

## Type of changes

- [x] Bugfix (non-breaking change which fixes an issue) (link the issue on the right)
- [ ] Documentation Update 
- [ ] Other (explain)

## Further comments
This could possibly go straight to `main`.


klevdiamanti:
There is two issues with OA leading workflows to fail:
- linting workflow needs a small addition.
- One of the dependencies, msigdbr, has undergone significant updates.
We are currently working on both. Hopefully, I'll have PRs to fix both by EOD.

OlaCaster:
> There is two issues with OA leading workflows to fail:
> 
> * linting workflow needs a small addition.
> * One of the dependencies, msigdbr, has undergone significant updates.
> 
> We are currently working on both. Hopefully, I'll have PRs to fix both by EOD.

Yes I guessed as much based on the other PRs in the list - so I wasn't particularly worried about the failing CI here 😄 
The `devtools` check passes for me locally at least.

klevdiamanti (approved):
LGTM

PR #551, opened 2025-04-11
Accept LODNPX as a column for normalization

klevdiamanti:
# Title: Accept LODNPX as a column for normalization
**Problem:** With NPXS v2.0 a new column, LODNPX, was introduced. The normalization function was not recognizing this as an alternative LOD column.

**Solution:** Added LODNPX to the alternative names of LOD.

## Checklist

## Type of changes

- [ ] Documentation Update 
- [ ] Other (explain)


MasoumehSheikhi:
test_that olink_bridgeability_plot  works failed on my Mac but all checked have passed here..?

klevdiamanti:
> test_that olink_bridgeability_plot  works failed on my Mac but all checked have passed here..?
> 
> 
> 

I have the same issue locally as vdiffr identifies differences between the plots; however, differences are caused by the OS's libraries. In this PR the code for tests of olink_bridgeability_plot was not modified.

MasoumehSheikhi:
As a result of binding rows, the output has both columns LOD and LODNPX with non-NA values for respective projects. Would it make sense to reorder the output columns so LOD and LODNPX are next to each other in the result? Or not a big deal?

MasoumehSheikhi (commented):
My two comments concludes my review.

MasoumehSheikhi (approved):
Approving the changes.

klevdiamanti:
> As a result of binding rows, the output has both columns LOD and LODNPX with non-NA values for respective projects. Would it make sense to reorder the output columns so LOD and LODNPX are next to each other in the result? Or not a big deal? 

This happens in the test because we're introducing the new column during the test. In reality, the order of the columns will not change withing the normalization function, and the original order of column will be retained.

PR #603, opened 2025-08-11
Bridging output format function

dtopouza:
# Title: Bridging format function
**Problem:** Update the bridging format option to include within-product bridging and comply with internal requirements for bridging output

**Key Features:**

* The _olink_normalization_product_format_ function is renamed to _olink_normalization_format_ and moved to olink_normalization_utils.R  to reflect change in scope
* Default output of _olink_normalization_ is **unchanged**
* _olink_norm_input_check_ now returns an additional list item named _non_overlapping_oid_ that lists all non-overlapping OlinkIDs between the two projects to be bridged
* If the optional _format == TRUE_ argument of _olink_normalization_ is used, and the normalization type is within- or cross-product bridging, theI function will edit the output to prepare for downstream analysis
* The formatted output includes non-overlapping and/or non-bridgeable assays in output with an adjustment factor of 0. A message informs the user of the number of assays returned without adjustment. 
* The formatted output removes negative controls and plate controls, and includes sample controls. A message informs the user of the number of control samples removed and lists their sample IDs.
* For cross-product bridging, the formatted output replaces the NPX values of adjusted proteins by the median-centered or QS-normalized NPX value, based on the recommendation.
* New unit tests added

## Checklist
- [x] Add neccessary documentation (if applicable)

## Type of changes

- [x] Code style update (formatting, renaming)
- [ ] Documentation Update 
- [ ] Other (explain)


klevdiamanti (changes requested):
I have made a series of suggestions that will allow another round of review on the format function itself. Can you please incorporate the suggested changes and then I can take a detailed view on the format function.

I have a major question though: shouldn't this function support all normalizations? Specifically:
1. Cross-product normalization is a special case, but parts of it are shared with other normalizations. This has been addressed as a separate case here, and this is the correct approach.
2. Bridge and subset normalization should undergo the same formatting before the output, except from the bridge samples which are not relevant for the subset normalization.
3. Reference median normalization should also output all assays from the input dataset, but not from the ones unique to the set of reference medians.

dtopouza:
> Shall we close this PR?

Yes, closing this PR as PR #637 supercedes it

PR #658, opened 2026-01-15
fixed regex and added test

kathy-nevola:
DarID regex issue that was triggering warning unnecessarily (ex: for Reveal DARIDs)


boxizhang (approved):
All good! Just one question, have we encountered this issue before? We should have the same problem with using "||", even for other products, right?

PR #761, opened 2026-05-28
Fix bugs and improve handling of annotation colors for pheatmap

### LINKED ISSUE #758: [BUG] olink_heatmap_plot no longer takes a list of annotation_colors
OlinkAnalyze_5.0.0 : **olink_heatmap_plot()** no longer passes  **annotation_colors** argument to pheatmap properly. 
Previous versions worked ok (OlinkAnalyze_4.5.0)

Adding colors with annotation_colors argument (named list of colors, as specificied in pheatmap help) gives following error:

```
Error in as.data.frame.default(x[[i]], optional = TRUE) : 
  cannot coerce class ‘"function"’ to a data.frame
```

klevdiamanti:
This pull request enhances the flexibility and robustness of heatmap annotation coloring in the `OlinkAnalyze` package, particularly around how annotation colors are generated, passed, and tested for heatmap plots. The changes improve both the internal logic for assigning annotation colors and the comprehensiveness of the test suite.

**Improvements to annotation color handling:**

* The `plot_heatmap_helper.R` logic now explicitly manages the `annotation_colors` parameter, ensuring that annotation colors are properly set to either user-specified values or a default (`NA`) when not provided. The internal argument `annot_col_int` is now consistently converted or removed as needed.
* In the color assignment process, the code now ensures that the color vectors are named according to the unique, sorted annotation values, which improves the reliability and clarity of color mapping in the heatmaps.

**Enhancements to test coverage and clarity:**

* The test suite for heatmap plotting (`test-plot_heatmap.R`) has been significantly expanded to cover multiple scenarios, including: single and multiple annotations, both auto-generated and manual color scales, and combinations thereof. This ensures that the new annotation color logic is robustly tested.
* Tests for helper functions have been updated to use the `annotation_colors` argument instead of the now-internal `annot_col_int`, and to provide named color vectors for clarity and correctness.
* All relevant tests now explicitly set `annotation_colors = NA` where appropriate, aligning with the updated function logic and ensuring consistent test behavior. [[1]](diffhunk://#diff-013dfd72fb295486a1bd36f30e0cfc4764443f9f64e66f19d76045f104b4dad7L238-R239) [[2]](diffhunk://#diff-013dfd72fb295486a1bd36f30e0cfc4764443f9f64e66f19d76045f104b4dad7L326-R330)

These changes collectively make the heatmap plotting functions more predictable and easier to use, while also ensuring that future changes are well-guarded by comprehensive tests.

This PR fixes issue #758.


OlinkAnalyze/R/plot_heatmap_helper.R
```diff
@@ -202,6 +202,16 @@ plot_heatmap_pheatmap_args <- function(df_wide,
     )
   }
 
+  # Add annotation colors to function call
+  if (is.null(pheatmap_args[["annot_col_int"]])) {
+    # NA is the default no-specification expected by pheatmap
+    pheatmap_args[["annotation_colors"]] <- NA
```
dtopouza, line 208: There is some odd behaviour where there are no cell borders when an annotation is used, but they show up when there is no annotation. This is a pheatmap issue, borders are by default on but get switched off when the heatmap gets past a certain size. I wonder if we should set border_color = NA as a default, for consistency.

Image

Image

OlinkAnalyze/R/plot_heatmap_helper.R
```diff
@@ -318,18 +328,27 @@ pheatmap_color_heatmap <- function(df,
...
+                                    function(x) length(unique(x))))
+
+    rep_num <- rep(names(variable_list), times = variable_list)
 
-  color_assignments <- with(data.frame(variable = rep_num, color = colors),
-                            split(color, variable))
+    color_assignments <- with(data.frame(variable = rep_num, color = colors),
+                              split(color, variable))
+
+    color_assignments <- lapply(vars_to_color, function(x) {
```
dtopouza, line 341: pheatmap_color_heatmap and some other helper functions (pheatmap_set_plot_theme, pheatmap_lst_styling) are a bit complicated and could really benefit from some commenting, to make them easier to understand and troubleshoot.

OlinkAnalyze/R/plot_heatmap_helper.R
```diff
@@ -209,6 +209,7 @@ plot_heatmap_pheatmap_args <- function(df_wide,
   } else {
     # Use the created list
     pheatmap_args[["annotation_colors"]] <- pheatmap_args[["annot_col_int"]]
+    pheatmap_args[["annot_col_int"]] <- NULL
```
dtopouza, line 212: Can we use annotation_colors throughout the function, instead of creating and deleting annot_col_int?

dtopouza (commented):
Looks good overall, had some suggestions in comments.

OlinkAnalyze/R/plot_heatmap_helper.R
klevdiamanti: `annotation_colors` can be a mixture of colors preset by the customer through the argument `annotation_colors` of the `olink_heatmap_plot()` function, and unset colors set via `variable_row_list`. This makes it tricky to keep everything in one list. `annot_col_int` functions as a useful intermediary.

OlinkAnalyze/R/plot_heatmap_helper.R
klevdiamanti: I totally agree that the function needs some good cleanup; note that it was all one endless function that we broke down to smaller ones but still needs work. On the other hand, getting this function cleaned up will not be quick and is sort of out of scope of this PR which aims to fix the underlying bug.

OlinkAnalyze/R/plot_heatmap_helper.R
klevdiamanti: I don't see this behavior; however, I never get the grid even when I set `border_color = "black"` which is very strange. Same behavior in both OA v4.5.0 and v5.0.0. Tested it both on PW (Ubuntu) and my local machine (Windows). Can it be difference in R version? Not sure what is going on. I'll make a card about this.

OlinkAnalyze/R/plot_heatmap_helper.R
dtopouza: that makes sense!

OlinkAnalyze/R/plot_heatmap_helper.R
dtopouza: Sounds good, and probably out of scope for this PR. I can add the code to replicate this

dtopouza (approved):
LGTM

PR #765, opened 2026-06-12
Enhance Olink dataset handling with check_log integration

klevdiamanti:
This pull request introduces several improvements and code quality updates to the OlinkAnalyze package. The most significant changes include expanded test coverage, improved S3 method integration for custom classes, and a series of code quality and documentation enhancements, especially for internal helper functions. The following summarizes the most important changes:

**Test Coverage and Imports:**
* Expanded the `Config/testthat/start-first` field in `DESCRIPTION` to include a broader set of functions for prioritized testing, improving test coverage for critical workflows.
* Added `pillar` to the `Imports` in `DESCRIPTION` to support new S3 method implementations.

**S3 Methods and Class Integration:**
* Registered new S3 methods for the `olink_class` class with `dplyr`, `pillar`, and `tibble`, improving compatibility with the tidyverse ecosystem.
* Exported new helper functions `olink_check_log` and `strip_check_log` for user accessibility. [[1]](diffhunk://#diff-0d79c569ac4cf769040f1850a507a1f57c41be705c7e26d1fab37c8551d49ed8R14) [[2]](diffhunk://#diff-0d79c569ac4cf769040f1850a507a1f57c41be705c7e26d1fab37c8551d49ed8R54-R57)
* Imported `tbl_sum` from `pillar` to support custom table summaries.

**Code Quality and Linting:**
* Updated numerous inline comments to use the explicit `# nolint: <rule>` syntax for clarity and to satisfy linters, especially in `Olink_one_non_parametric.R`. [[1]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L98-R96) [[2]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L111-R110) [[3]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L131-R134) [[4]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L170-R166) [[5]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L191-R204) [[6]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L264-R260) [[7]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L298-R300) [[8]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L331-R327) [[9]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L432-R423) [[10]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L461-R454) [[11]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L489-R500) [[12]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L566-R557) [[13]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L592-R584) [[14]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L632-R623) [[15]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L684-R675) [[16]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L719-R710)
* Changed calls from `run_check_npx` to `get_check_npx` for improved naming consistency and clarity. [[1]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L98-R96) [[2]](diffhunk://#diff-31914759813a7eb62fcac1211dc44000a98cd1996bf9be4923f1c606873b6ac1L461-R454)

**Documentation Improvements:**
* Added `@noRd` to internal helper functions and removed `seealso` from `@inherit` tags for cleaner internal documentation in files such as `check_boolean.R`, `check_character.R`, and `check_columns.R`. [[1]](diffhunk://#diff-82e65d47c14c0de9b8269fcce95e97727325409178480f23338168b7d8113087L3-R6) [[2]](diffhunk://#diff-82e65d47c14c0de9b8269fcce95e97727325409178480f23338168b7d8113087L40-R44) [[3]](diffhunk://#diff-64f907635b3782ad9afc7ea31fa9b770b70709e941782c51996d65722ba1737eL3-R6) [[4]](diffhunk://#diff-64f907635b3782ad9afc7ea31fa9b770b70709e941782c51996d65722ba1737eL44-R48) [[5]](diffhunk://#diff-beec72c852a36fbbf47f14f65a87eceeb32673a2897a63379c90acc08662ee03R70) [[6]](diffhunk://#diff-beec72c852a36fbbf47f14f65a87eceeb32673a2897a63379c90acc08662ee03R173)

**Build and Ignore Files:**
* Updated `.Rbuildignore` to exclude `.positai` and `.claude` files from builds.

These changes collectively enhance package robustness, maintainability, and integration with the tidyverse, while improving internal code quality and documentation.


OlinkAnalyze/R/check_npx.R
kathy-nevola, line 1: should check_NPX return an olink class with the check log?

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+#'
+#' @return An object of class `olink_class`, which inherits from
+#' `tbl_df`, `tbl`, and `data.frame`, with the check log stored as an
+#' attribute.
+#'
+#' @seealso [olink_check_log()] for retrieving the check log from an object.
+#'
+#' @keywords internal
+#'
+new_olink_class <- function(df,
```
kathy-nevola, line 53: if new_olink_class is used on an existing olink class, then maybe a warning should be thrown?

kathy-nevola:
how is the workflow from the user's side expected to work? if a check_log is being generated on read_npx and then a user runs check_npx and provides the check_log to the function with the olink_class data then the function will use the attached check log over the user provided check log. However the workflow changes we will need to update in the overview vignette. Specifically for the case of using quant data instead of NPX in downstream analysis (as this question has already come up a few times).

kathy-nevola:
should npx_data1 and npx_data2 be converted to olink_class instead?

OlinkAnalyze/R/check_npx.R
```diff
@@ -180,169 +180,134 @@ check_npx <- function(df,
...
+  different_content <- shared_names[
+    vapply(
+      shared_names,
+      function(nm) {
+        !same_content( # nolint: return_linter
+          x = check_log_names[[nm]],
+          y = check_log_local_names[[nm]]
+        )
+      },
+      logical(1)
```
kathy-nevola, line 296: When preferred name is different from default get_preferred names creates a named vector, else this returns NULL. This is just used to conserve any preferred names as specified by the user.

OlinkAnalyze/R/clean_npx.R
```diff
@@ -89,23 +89,22 @@
...
 #' )
+#'
+#' # run clean_npx but keep duplicated samples
+#' OlinkAnalyze::clean_npx(
+#'   df = OlinkAnalyze::npx_data1,
+#'   remove_dup_sample_id = FALSE
+#' )
 #' }
 #'
 clean_npx <- function(df,
```
kathy-nevola, line 110: Note to self: This is an overview of the workflow 

Image

OlinkAnalyze/R/olink_boxplot.R
```diff
@@ -103,17 +105,15 @@ olink_boxplot <- function(df,
...
     df = df,
-    check_log = check_log,
+    out_df = "tibble",
+    check_log = get_check_npx(df = df, check_log = check_log),
     remove_qc_warning = FALSE,
     remove_assay_warning = FALSE,
     verbose = FALSE
   )
-  check_log_clean <- run_check_npx(df = df, check_log = NULL) |>
-    suppressMessages() |>
```
kathy-nevola, line 115: ~~do we not need these anymore because we assume clean NPX will take care of all warnings and messages~~

clean_npx will regenerate check_log at the end of check_npx call, this will just extract it from the olink_class

OlinkAnalyze/R/olink_normalization_product.R
kathy-nevola, line 109: add .noRd and keyword internal since it runs internally

OlinkAnalyze/R/olink_normalization_product.R
```diff
@@ -30,9 +30,9 @@
 #' \code{\link{olink_norm_input_check}} when it is called from
 #' \code{\link{olink_normalization}}. In addition,
 #' \code{\link{olink_normalization}} has also utilized
-#' \code{\link{norm_internal_rename_cols}} to rename the columns of the
-#' non-reference dataset according to the ones of the reference dataset, hence
-#' all column names should match.
+#' `norm_internal_rename_cols` to rename the columns of the non-reference
```
kathy-nevola: should this be 'norm_internal_preferred_names' instead

OlinkAnalyze/R/olink_normalization.R
```diff
@@ -277,6 +258,19 @@ olink_normalization <- function(df1,
     reference_medians = reference_medians
   )
 
+  # kill switch
```
kathy-nevola, line 261: is this a temporary solution?

OlinkAnalyze/DESCRIPTION
```diff
@@ -68,7 +68,7 @@ URL:
     https://7074596.fs1.hubspotusercontent-na1.net/hubfs/7074596/01-User%20Manuals%20for%20website/1494-Olink-Analyze-Cheatsheet.pdf
 Config/testthat/edition: 3
 Config/testthat/parallel: true
-Config/testthat/start-first: read_npx_l*, read_npx_w*
+Config/testthat/start-first: olink_pathway_enrichment, olink_normalization, read_npx_legacy, read_npx_wide, olink_normalization_utils, olink_lmer, olink_ordinal_regression, olink_normalization_n, olink_normalization_format, pca_plot, read_npx_format, plot_is_bridgeable, clean_npx, plot_heatmap
```
kathy-nevola, line 71: I dont think this is an issue, but I dont understand why these start first as they seem like the ones that would take the longest to run. Would we instead want the short and quick ones to run first so that if they fail the workflow will stop?

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+#'     check_log = check_log
+#'   )
+#'
+#'   # retrieve check_log
+#'   OlinkAnalyze::olink_check_log(
+#'     df = npx_obj
+#'   )
+#' }
+#'
+olink_check_log <- function(df) {
```
kathy-nevola: recommend using a name like olink_extract_check_log to make it clear that we are extracting the check log, not generating it?

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+#' ArrowObject, returning a plain Arrow table without any attached check log.
+#'
+#' @param df An ArrowObject (e.g. `arrow::Table` or `arrow::Dataset`).
+#'
+#' @return The ArrowObject with the `olink_check_log` metadata removed.
+#'
+#' @keywords internal
+#' @noRd
+#'
+strip_check_log_arrow <- function(df) {
```
kathy-nevola: maybe rm_check_log_arrow might be a better name? but it is also fine as is

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+#'
+#' The `olink_class` class survives most dplyr operations (e.g.
+#' [`dplyr::filter()`], [`dplyr::mutate()`], [`dplyr::select()`]) through the
+#' [`dplyr::dplyr_reconstruct()`] mechanism. Note that operations that combine
+#' multiple data frames (e.g. [`dplyr::bind_rows()`]) may strip the class and
+#' check log.
+#'
+#' @name olink_class
+#' @keywords internal
+#'
```
kathy-nevola, line 29: do we want Rd generated externally for this class?

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+  if (inherits(x = df, what = "olink_class")) {
+
+    return(attr(x = df, which = "check_log", exact = TRUE))
+
+  } else if (check_is_arrow_object(x = df, error = FALSE)) {
+
+    check_log_encoded <- df$metadata[["olink_check_log"]]
+
+    if (is.null(check_log_encoded)) {
+      return(NULL)
```
kathy-nevola, line 297: should it error on a NULL response?

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+#' @noRd
+#'
+strip_check_log_arrow <- function(df) {
+
+  check_is_arrow_object(x = df, error = TRUE)
+
+  existing_metadata <- df$metadata
+
+  if ("olink_check_log" %in% names(existing_metadata)) {
+    existing_metadata[["olink_check_log"]] <- NULL
```
kathy-nevola, line 359: if olink_check_log is the only metadata this will still leave an empty named list in the metadata. However in cases where metadata is present this works as expected

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+      }
+    }, # nolint: return_linter
+    error = function(e) {
+      NULL # nolint: return_linter
+    }
+  )
+
+  cli::cli_warn(
+    c(
+      "!" = "Could not deserialize {.arg check_log} from Arrow metadata.",
```
kathy-nevola, line 549: Are there any cases where we would expect check_log not to be able to be deserialized?

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+base64_encode <- function(raw_bytes) {
+
+  if (length(raw_bytes) == 0L) return("")
+
+  base64_chars <- c(LETTERS, letters, 0L:9L, "+", "/")
+
+  n <- length(raw_bytes)
+  int_vals <- as.integer(raw_bytes)
+
+  # pad to multiple of 3
```
kathy-nevola, line 581: I dont understand the math here but the functions work as expected

OlinkAnalyze/R/olink_class.R
```diff
@@ -0,0 +1,710 @@
...
+#'
+tbl_sum.olink_class <- function(x, ...) {
+
+  default_header <- NextMethod()
+
+  check_log <- attr(x = x, which = "check_log", exact = TRUE)
+
+  if (!is.null(check_log)) {
+    olink_class_name <- c(default_header,
+                          "Check log" = "attached")
```
kathy-nevola: recommend calling this olink_check_log or check_log to match arrow and function arguments

OlinkAnalyze/R/utils.R
```diff
@@ -36,13 +37,119 @@ remove_all_na_cols <- function(df) {
...
+                                    out_df) {
+
+  # check that out_df is ok
+  check_out_df_arg(out_df = out_df)
+
+  if (check_is_dataset(x = df, error = FALSE)) {
+
+    if (out_df == "tibble") {
+
+      return(dplyr::as_tibble(df))
```
kathy-nevola, line 98: should this convert to an olink_class and keep the meta-data? Or is that too complex. (in reference to normalization function use as kill switch)

OlinkAnalyze/R/olink_normalization.R
```diff
@@ -277,6 +258,19 @@ olink_normalization <- function(df1,
...
+  # convert to tibble if no time to get arrow work across the function
+  if (lst_check$ref_out_df == "arrow") {
+    lst_check$ref_df <- convert_read_npx_output(
```
kathy-nevola, line 264: is the plan to add this functionality at a different time or is not possible to work with arrow objects here. 

also not that check_log is dropped here

OlinkAnalyze/R/olink_normalization.R
```diff
@@ -277,6 +258,19 @@ olink_normalization <- function(df1,
     reference_medians = reference_medians
   )
```
kathy-nevola, line 260: should check_logs be extracted from df1 and df2 to use in olink_norm_input_check

OlinkAnalyze/R/olink_normalization.R
```diff
@@ -381,82 +395,110 @@ olink_normalization <- function(df1,
 
   }
```
kathy-nevola, line 414: in cases where there is a diff_preferred_column, I would expect the columns with different names to be combined (ex: QC_Warning and SampleQC) rather than just have information for one project or the other. Optionally the column from the non-reference project could be removed or there could be two duplicate columns (QC_Warning and SampleQC) with the preferred column indicated in check_log

OlinkAnalyze/R/olink_normalization.R
```diff
@@ -381,82 +395,110 @@ olink_normalization <- function(df1,
...
+    ) |>
+    dplyr::filter(
+      .data[["ref_col_name"]] != .data[["not_ref_col_name"]]
+    ) |>
+    dplyr::select(
+      dplyr::all_of(
+        c("ref_col_key", "ref_col_name")
+      )
+    ) |>
+    tibble::deframe()
```
kathy-nevola, line 516: What is the use case for the preferred names downstream. For example: if you had QC_Warning in the ref dataset and SampleQC in the other, then the qc_warning preferred key would be "QC_Warning" but the SampleQC data would not be mapped to the QC_Warning column. Additionally, only the ref can be used as the preferred names.

OlinkAnalyze/R/olink_normalization_utils.R
```diff
@@ -295,12 +300,16 @@ olink_norm_input_check <- function(df1,
     ref_samples = NULL,
     ref_name = NULL,
     ref_product = NULL,
+    ref_out_df = NULL,
```
kathy-nevola: This naming makes me assume it is the bridged data (output df). maybe format or output format would be a better name

klevdiamanti:
Test have started failing on results from ORA pathway enrichment and update of the R package `rstatix` that causes tests failing the Friedman post-hoc tests. The former is fixed in PR #773 and the latter is already fixed in this [branch](https://github.com/Olink-Proteomics/OlinkRPackage/tree/OA-2606-Update-reference-results-from-Friedman-test). Both PRs can be reviewed and merged once this PR #765 is reviewed and merged.

OlinkAnalyze/R/olink_normalization.R
```diff
@@ -290,19 +284,39 @@ olink_normalization <- function(df1,
...
+        df = lst_check$not_ref_original_df,
+        out_df = "tibble"
+      )
+    }
+
     ## rename non-reference columns to reference columns ----
 
-    # update selected colnames of not_ref_df based on colnames of ref_df
-    lst_check$not_ref_df <- norm_internal_rename_cols(
+    diff_preferred_names <- norm_internal_preferred_names(
```
kathy-nevola, line 313: In the previous version column names with the same dictionary key would be concatenated even if the column name was not the same (in some but not all cases, this didnt occur for different OlinkID names). Now both column names are kept and NA is filled in for the project where it wasnt present.

For example if df1 has SampleQC and df2 has QC_Warning and they were bridged together, previously the output data would just have SampleQC and would have the values from SampleQC and QC_Warning present in the SampleQC column. However now SampleQC and QC_Warning are both present in the outputs with NA as outputs for QC_Warning for df1 datapoints and NA as outputs for SampleQC warning for df2 datapoints.

OlinkAnalyze/tests/testthat/test-Olink_one_non_parametric.R
```diff
@@ -121,6 +214,16 @@ test_that(
       "dunn.test version = 1.4.1 contains a bug."
     )
 
+    # expected results ----
+
+    kruskal_results_nrow <- 184L
+    kruskal_results_ncol <- 11L
+
+    kruskal_posthoc_results_nrow <- 190L
```
kathy-nevola, line 222: kruskal_posthoc_results_nrow and kruskal_posthoc_results_ncol variables are not used. I think they should be added to row 313 and 321 and the value of kruskal_posthoc_results_ncol should be 10L. Additionally should the kruskal tests be repeated with the olink_class? (might be overkill)

OlinkAnalyze/tests/testthat/test-check_columns.R
```diff
@@ -74,6 +74,83 @@ test_that(
...
+
+    # SampleID and (F or E) exist -> error as neither E nor F exist
+    expect_error(
+      object = check_columns(df = npx_data1_obj,
+                             col_list = list("SampleID",
+                                             c("Count", "ExtNPX"))),
+      regexp = "is missing columns that should be present in at least one"
+    )
+
+    # SampleID and (Count or ExtNPX) exist -> error as neither E nor F exist
```
kathy-nevola:
```suggestion
    # SampleID and (Count or ExtNPX) and (PlateLOD or MaxLOD) exist -> error as neither (Count or ExtNPX) nor (PlateLOD or MaxLOD) exist
```

OlinkAnalyze/tests/testthat/test-check_df.R
```diff
@@ -579,6 +579,197 @@ test_that(
...
+    expect_error(
+      object = check_is_olink_class(x = data.frame(a = c(1, 2),
+                                                   b = c("a", "b"),
+                                                   c = c(TRUE, FALSE)) |>
+                                      arrow::as_arrow_table(),
+                                    error = TRUE),
+      regexp = gsub(pattern = " ",
+                    replacement = "([[:space:]].*|\\n.*)?",
+                    x = "is not an Olink class object!",
+                    fixed = TRUE)
```
kathy-nevola, line 675: additional test, expect error and/or false when regular NPX df is checked as olink_class 

`check_is_olink_class(x = npx_data1) `

OlinkAnalyze/tests/testthat/test-check_df.R
```diff
@@ -799,6 +990,64 @@ test_that(
...
+          "A" = c(1, 2.2, 3.14),
+          "B" = c("a", "b", "c"),
+          "C" = c(TRUE, TRUE, FALSE),
+          "D" = c("NA", "B", NA_character_),
+          "E" = c(1L, 2L, 3L)
+        )
+
+        sep_arrow <- ","
+
+        # write the coma-delimited file
```
kathy-nevola, line 1021:
```suggestion
        # write the comma-delimited file
```

OlinkAnalyze/tests/testthat/test-check_npx.R
```diff
@@ -368,191 +369,409 @@ test_that(
...
-    expect_warning(
-      object = run_check_npx(df = npx_data1,
-                             check_log = test_check_log),
-      regexp = "Unexpected key \"i_am_a_new_element\" corresponding to column"
-    )
+    obj <- new_olink_class(df = df_tbl,
+                           check_log = check_log_result)
+
+    # pass a deliberately different (invalid) check_log argument; the embedded
+    # one should be used instead without raising an error
```
kathy-nevola, line 648: If a user generates a new check_log and provides it to the function, would we want that to be used over the default one generated by read_npx on import?

OlinkAnalyze/tests/testthat/test-clean_npx.R
```diff
@@ -2361,11 +2659,472 @@ test_that(
...
+      ),
+      regexp = paste("Duplicate SampleIDs detected: \"CONTROL_SAMPLE_AS 1\"",
+                     "and \"CONTROL_SAMPLE_AS 2\"")
+    )
+
+    expect_identical(
+      object = olink_check_log(df = curr_result_v2)$col_names,
+      expected = check_log_tmp$col_names
+    )
+  }
```
kathy-nevola, line 3136: add test for if olink_class and check_log are both provided that correct log is used. (i.e same as in check_npx, https://github.com/Olink-Proteomics/OlinkRPackage/pull/765#discussion_r3545320889)

OlinkAnalyze/tests/testthat/test-dist_plot.R
```diff
@@ -121,3 +121,129 @@ test_that(
...
+    expect_no_warning(
+      object = expect_no_error(
+        object = expect_no_message(
+          object = distr_plot_obj <- olink_dist_plot(
+            df = npx_data1_filtered_obj
+          )
+        )
+      )
+    )
+
```
kathy-nevola, line 188:
```suggestion
# Dist plot using olink_class is same as dist plot using df and check_log
```

OlinkAnalyze/tests/testthat/test-dist_plot.R
```diff
@@ -121,3 +121,129 @@ test_that(
...
+    expect_no_warning(
+      object = expect_no_error(
+        object = expect_no_message(
+          object = distr_plot_arrow <- olink_dist_plot(
+            df = npx_data1_filtered_arrow
+          )
+        )
+      )
+    )
+
```
kathy-nevola, line 230:
```suggestion
# Dist plot using arrow object with check_log is same as dist plot with df and check_log
```

OlinkAnalyze/tests/testthat/test-dist_plot.R
```diff
@@ -121,3 +121,129 @@ test_that(
...
+      object = expect_no_error(
+        object = expect_no_message(
+          object = distr_plot_treat_obj <- olink_dist_plot(
+            df = npx_data1_filtered_obj,
+            color_g = "Treatment"
+          )
+        )
+      )
+    )
+
```
kathy-nevola, line 205:
```suggestion
# Dist plot using olink_class is same as dist plot using df and check_log when defining color_g
```

OlinkAnalyze/tests/testthat/test-dist_plot.R
```diff
@@ -121,3 +121,129 @@ test_that(
...
+      object = expect_no_error(
+        object = expect_no_message(
+          object = distr_plot_treat_arrow <- olink_dist_plot(
+            df = npx_data1_filtered_arrow,
+            color_g = "Treatment"
+          )
+        )
+      )
+    )
+
```
kathy-nevola, line 248:
```suggestion
# Dist plot using arrow object with check_log is same as dist plot with df and check_log and defining color_g
```

OlinkAnalyze/tests/testthat/test-olink_anova.R
```diff
@@ -155,6 +157,39 @@ test_that(
       object = anova_res_site,
       expected = reference_results$anova_site
     )
+
```
kathy-nevola, line 160: Should we also test if it works with the arrow object or is that part of the next stage of optimization?

OlinkAnalyze/tests/testthat/test-olink_boxplot.R
```diff
@@ -306,3 +314,201 @@ test_that(
     )
   }
 )
+
+test_that(
+  "olink_boxplot - works - vdiffr olink_class",
+  {
```
kathy-nevola, line 320: Is it worth adding the tests for time+site, time+posthoc, and changing color option? My assumption is if the figure is generated correctly once it *should* work in all cases, so it might be overkill.

OlinkAnalyze/tests/testthat/test-olink_class.R
```diff
@@ -0,0 +1,608 @@
...
+  "new_olink_class - works - creates an olink_class from tibble and check_log",
+  {
+    result <- new_olink_class(
+      df = npx_data1,
+      check_log = npx_data1_check_log
+    )
+
+    expect_s3_class(object = result, class = "olink_class")
+    expect_s3_class(object = result, class = "tbl_df")
+    expect_identical(object = dim(result), expected = c(29440L, 17L))
```
kathy-nevola: could make this dynamic instead of hard coding by expecting the same dimensions of npx_data1?

OlinkAnalyze/tests/testthat/test-olink_class.R
```diff
@@ -0,0 +1,608 @@
...
+    expect_null(object = olink_check_log(df = result))
+  }
+)
+
+test_that(
+  "strip_check_log - works - returns plain tibble unchanged",
+  {
+    result <- strip_check_log(df = npx_data1)
+
+    expect_s3_class(object = result, class = "tbl_df")
```
kathy-nevola, line 569: add test that s3 class is not olink_class

OlinkAnalyze/tests/testthat/test-olink_class.R
```diff
@@ -0,0 +1,608 @@
...
+  }
+)
+
+test_that(
+  "strip_check_log - works - returns ArrowObject without metadata unchanged",
+  {
+    arrow_tbl <- arrow::as_arrow_table(x = npx_data1)
+
+    result <- strip_check_log(df = arrow_tbl)
+
```
kathy-nevola, line 581: could add metadata to arrow table (as would be in a parquet) and make sure metadata is preserved

OlinkAnalyze/tests/testthat/test-olink_lmer.R
```diff
@@ -58,6 +60,63 @@ test_that(
       expected = reference_results$lmer,
       tolerance = 1e-4
     )
+
+    # olink_class ----
+
```
kathy-nevola, line 65: do we expect this to work with the arrow data-type as well?

OlinkAnalyze/tests/testthat/test-read_npx_zip.R
```diff
@@ -403,18 +444,7 @@ test_that(
...
-                  object = read_npx_zip(file = zfile_test,
-                                        out_df = "arrow",
-                                        long_format = TRUE,
-                                        .ignore_files =
-                                          c(basename(readmefile_test),
-                                            basename(rfile_test)),
-                                        quiet = TRUE),
-                  regexp = "Unable to confirm the \"long\" format from the inpu"
-                )
+                expect_no_error
```
kathy-nevola: this is incomplete

OlinkAnalyze/tests/testthat/helper-check.R
```diff
@@ -20,3 +20,43 @@ check_snap_exist <- function(test_dir_name, snap_name) {
...
+  testthat::expect_equal(
+    object$mapping,
+    expected$mapping
+  )
+
+  testthat::expect_equal(
+    lapply(object$layers, function(x) class(x$geom)[1L]),
+    lapply(expected$layers, function(x) class(x$geom)[1L])
+  )
+
```
kathy-nevola, line 50: could also add position, scales, and facet for extra safety

```suggestion
testthat::expect_equal(
    lapply(object$layers, function(x) class(x$position)[1L]),
    lapply(expected$layers, function(x) class(x$position)[1L])
  )
```

```suggestion
testthat::expect_equal(
    object$scales,
    expected$scales
  )
```

```suggestion
testthat::expect_equal(
    object$facet,
    expected$facet
  )
```

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -480,186 +664,266 @@ test_that(
...
+    expect_s3_class(object = bridge_no_norm_arrow, class = "olink_class")
+    expect_s3_class(object = bridge_no_norm_arrow, class = "tbl_df")
+
     expect_equal(
-      object = intensity_norm,
-      expected = ref_norm_res$lst_norm$intensity_norm$norm,
+      object = strip_check_log(df = bridge_no_norm_arrow),
+      expected = ref_bridge_norm_v1,
       tolerance = 1e-4
     )
```
kathy-nevola, line 862: same as above, should we test with norm col, with non-overlapping assays, with ext controls and with non-overlapping assays and external contols

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -709,399 +970,589 @@ test_that(
...
+    expect_s3_class(object = intensity_no_norm_arrow, class = "olink_class")
+    expect_s3_class(object = intensity_no_norm_arrow, class = "tbl_df")
+
     expect_equal(
-      object = subset_norm_v1,
-      expected = subset_int_norm_v1,
+      object = strip_check_log(df = intensity_no_norm_arrow),
+      expected = ref_norm_res$lst_norm$intensity_norm$no_norm,
       tolerance = 1e-4
     )
```
kathy-nevola, line 1163: same as above (norm, no lod, multi-lod)

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -1179,116 +1701,370 @@ test_that(
...
+                     "\"CONTROL_SAMPLE_AS 2\"")
+    )
+
+    expect_s3_class(object = subset_norm_v1_arrow, class = "olink_class")
+    expect_s3_class(object = subset_norm_v1_arrow, class = "tbl_df")
+
+    expect_equal(
+      object = strip_check_log(df = subset_norm_v1_arrow),
+      expected = subset_int_norm_v1,
+      tolerance = 1e-4
```
kathy-nevola, line 1988: Same as above. Should we test for w/o LOD or with multi-lod?

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -18,21 +18,65 @@ test_that(
 
     ### bridge normalization - no norm column ----
 
-    expect_message(
-      object = expect_warning(
+    expect_warning(
```
kathy-nevola, line 24: norm_internal_rename_cols (line 4) isnt a function anymore

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -41,122 +85,241 @@ test_that(
...
-      regexp = "Output includes two sets of bridging samples"
+      regexp = paste("Duplicate SampleIDs detected: \"A13\", \"A29\", \"A30\",",
+                     "\"A36\", \"A45\", \"A46\", \"A52\", \"A63\", \"A71\",",
+                     "\"A73\", \"B3\", \"B37\", \"B4\", \"B45\", \"B63\",",
+                     "\"B75\", \"CONTROL_SAMPLE_AS 1\", and",
+                     "\"CONTROL_SAMPLE_AS 2\"")
     )
 
+    expect_s3_class(object = bridge_no_norm_arrow, class = "olink_class")
+    expect_s3_class(object = bridge_no_norm_arrow, class = "tbl_df")
```
kathy-nevola, line 321: should we test other cases (norm, no lod, multi-lod)

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -480,186 +664,266 @@ test_that(
...
-        regexp = "Subset normalization will be performed!"
+        object = expect_message(
+          object = bridge_no_norm_obj <- olink_normalization(
+            df1 = df_norm_ref_v1_obj,
+            df2 = df_norm_not_ref_v1_obj,
+            overlapping_samples_df1 = ref_norm_res$lst_sample$bridge_samples,
+            df1_project_nr = "df1_norm",
+            df2_project_nr = "df2_norm",
+            reference_project = "df1_norm",
+            format = FALSE
```
kathy-nevola:
```suggestion
            format = TRUE
```

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -480,186 +664,266 @@ test_that(
...
+      object = expect_message(
+        object = expect_message(
+          object = bridge_no_norm_arrow <- olink_normalization(
+            df1 = df_norm_ref_v1_arrow,
+            df2 = df_norm_not_ref_v1_arrow,
+            overlapping_samples_df1 = ref_norm_res$lst_sample$bridge_samples,
+            df1_project_nr = "df1_norm",
+            df2_project_nr = "df2_norm",
+            reference_project = "df1_norm",
+            format = FALSE
```
kathy-nevola:
```suggestion
            format = TRUE
```

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -709,399 +970,589 @@ test_that(
...
+                     "\"CONTROL_SAMPLE_AS 2\"")
+    )
+
+    expect_s3_class(object = subset_no_norm_arrow, class = "olink_class")
+    expect_s3_class(object = subset_no_norm_arrow, class = "tbl_df")
+
+    expect_equal(
+      object = strip_check_log(df = subset_no_norm_arrow),
+      expected = ref_norm_res$lst_norm$subset_norm$no_norm,
+      tolerance = 1e-4
```
kathy-nevola, line 1462: same as above (norm, no lod, multi-lod)

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -1297,14 +2073,141 @@ test_that(
...
+                       "named \"Normalization\"")
+      ),
+      regexp = paste("Duplicate SampleIDs detected: \"CONTROL_SAMPLE_AS 1\"",
+                     "and \"CONTROL_SAMPLE_AS 2\"")
+    )
+
+    expect_s3_class(object = ref_med_no_norm_arrow, class = "olink_class")
+    expect_s3_class(object = ref_med_no_norm_arrow, class = "tbl_df")
+
+    expect_equal(
```
kathy-nevola, line 2209: same as above

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -1571,241 +2486,1639 @@ test_that(
...
+      ),
+      regexp = paste("Duplicate SampleIDs detected: \"CONTROL_SAMPLE_AS 1\"",
+                     "and \"CONTROL_SAMPLE_AS 2\"")
+    )
+
+    expect_s3_class(object = refmed_norm_v1_arrow, class = "olink_class")
+    expect_s3_class(object = refmed_norm_v1_arrow, class = "tbl_df")
+
+    expect_equal(
+      object = strip_check_log(df = refmed_norm_v1_arrow),
```
kathy-nevola: same as above, non-overlap, ext control, non-overlap + ext control

OlinkAnalyze/tests/testthat/test-olink_normalization.R
kathy-nevola: also norm_internal_preferred_names is tested on line 4568 so this comment might be outdated

OlinkAnalyze/tests/testthat/test-olink_normalization.R
```diff
@@ -2278,19 +4689,15 @@ test_that(
...
-      names()
+      not_ref_cols = not_ref_df_check$col_names
+    )
 
     # ensure that
     expect_identical(
       object = update_cnames_v1,
-      expected = c(names(npx_data2), c("Normalization")) |>
-        stringr::str_replace("QC_Warning", "SampleQC")
+      expected = c("panel_version" = "Panel_Version", "qc_warning" = "SampleQC")
```
kathy-nevola, line 4703: should add test to check concatenation of columns with same key

OlinkAnalyze/tests/testthat/test-olink_normalization_n.R
```diff
@@ -1004,151 +1154,318 @@ test_that(
...
+    )
+
+    expect_s3_class(object = intensity_no_norm_arrow, class = "olink_class")
+    expect_s3_class(object = intensity_no_norm_arrow, class = "tbl_df")
+
+    expect_equal(
+      object = strip_check_log(df = intensity_no_norm_arrow),
+      expected = ref_norm_res$lst_norm$intensity_norm$no_norm,
+      tolerance = 1e-4
+    )
```
kathy-nevola, line 1349: same comments as in olink_normalization.R with additional columns

OlinkAnalyze/tests/testthat/test-olink_ttest.R
```diff
@@ -33,6 +35,42 @@ test_that(
...
+        ),
+        regexp = "Variable converted from character to factor: Treatment"
+      ),
+      regexp = "T-test is performed on Treated - Untreated."
+    )
+
+    expect_equal(
+      object = t_test_res_obj,
+      expected = reference_results$t_test
+    )
```
kathy-nevola, line 73: I assume we arent testing this on arrow because it hasnt been optimized for arrow yet?

OlinkAnalyze/R/pca_plot.R
```diff
@@ -81,19 +81,15 @@
 #'              ignore.case = TRUE)
```
kathy-nevola, line 81: on line 16, the % are not rendered correctly in the documentation and reads as follow:

```
Imputation by the median is done for assays with missingness <10\ for multi-plate projects and <5\
```

OlinkAnalyze/tests/testthat/test-utils.R
```diff
@@ -132,6 +132,586 @@ test_that(
...
+)
+
+# Test convert_read_npx_output ----
+
+test_that(
+  "convert_read_npx_output - works - df: arrow; out: tibble",
+  {
+    # simple arrow ----
+
+    expect_true(
```
kathy-nevola: could we do this with expect_s3_class

OlinkAnalyze/tests/testthat/test-utils.R
```diff
@@ -132,6 +132,586 @@ test_that(
...
+    )
+
+    expect_no_condition(
+      object = npx_data1_tibble <- convert_read_npx_output(
+        df = npx_data1_arrow,
+        out_df = "tibble"
+      )
+    )
+
+    expect_true(object = inherits(x = npx_data1_tibble, what = "tbl_df"))
```
kathy-nevola: same as above: expect_s3_class?

OlinkAnalyze/tests/testthat/test-utils.R
```diff
@@ -132,6 +132,586 @@ test_that(
...
+    )
+
+    expect_no_condition(
+      object = npx_data1_arrow_v2 <- convert_read_npx_output(
+        df = npx_data1_arrow,
+        out_df = "arrow"
+      )
+    )
+
+    expect_true(object = inherits(x = npx_data1_arrow_v2, what = "ArrowObject"))
```
kathy-nevola: can also use expect_s3_class here

OlinkAnalyze/tests/testthat/test-utils.R
kathy-nevola: this is true throughout this file/section wherever inherits is used (except the expect_false statments)

OlinkAnalyze/tests/testthat/test-utils.R
```diff
@@ -263,13 +966,86 @@ test_that(
...
+        object = expect_no_message(
+          object = osi_data_obj <- attach_check_log(
+            df = osi_data,
+            out_df = "tibble"
+          )
+        )
+      )
+    )
+
+    # OSISUmmary ----
```
kathy-nevola, line 984:
```suggestion
    # OSISummary ----
```

OlinkAnalyze/tests/testthat/test-plot_point_range.R
```diff
@@ -107,6 +154,50 @@ test_that(
...
+            df = npx_data1,
+            out_df = "tibble"
+          ),
+          regexp = paste("Duplicate SampleIDs detected: \"CONTROL_SAMPLE_AS",
+                         "1\" and \"CONTROL_SAMPLE_AS 2\"")
+        )
+      )
+    )
+
+    expect_no_error(
```
kathy-nevola, line 173: we arent testing with arrow here because this runs olink_lmer under the hood which doesnt support arrow yet

OlinkAnalyze/tests/testthat/test-plot_is_bridgeable.R
```diff
@@ -102,49 +124,70 @@ test_that(
...
+    )
 
     overlapping_samples <- intersect(
       x = npx_ht$SampleID,
       y = npx_3072$SampleID
     )
 
-    expect_message(
+    expect_warning(
       object = expect_message(
```
kathy-nevola, line 165: should we also have the expect_equal_ggplot here?

OlinkAnalyze/tests/testthat/test-plot_is_bridgeable.R
```diff
@@ -781,66 +945,94 @@ test_that(
...
+                    df1 = npx_ht,
+                    df2 = npx_3072,
+                    overlapping_samples_df1 = overlapping_samples,
+                    df1_project_nr = "Explore HT",
+                    df2_project_nr = "Explore 3072",
+                    reference_project = "Explore HT",
+                    # setting format = TRUE to test that function works for when
+                    # BridgingRecommendation is "NotOverlapping"
+                    format = TRUE,
+                    df1_check_log = check_npx(df = npx_ht) |>
```
kathy-nevola: can remove check_logs here

OlinkAnalyze/tests/testthat/test-plot_is_bridgeable.R
```diff
@@ -877,70 +1069,96 @@ test_that(
...
```
kathy-nevola: can remove check_logs here?

OlinkAnalyze/tests/testthat/test-plot_heatmap.R
```diff
@@ -221,3 +221,146 @@ test_that(
     })
   }
 )
+
+test_that(
```
kathy-nevola, line 225: we cant use expect_equal_ggplot here because heatmap function ggplot object does not have informative data

kathy-nevola (changes requested):
A few changes and questions. One thing to confirm is that we aren't testing the arrow objects on the stats functions. Is this because those functions havent been updated yet. 

Biggest requested change is preferred_names function which does not have the same output as the rename_cols (specifically with QC_Warning and SampleQC differing between dfs which now results in 2 columns instead of 1 combined column)

OlinkAnalyze/DESCRIPTION
klevdiamanti: My thinking is that, while working on a feature, you should only run the tests that are affected by that feature rather than the full test suite. That way, developers are not delayed by slower tests unless they are working on functionality that those tests cover. If something breaks, it should ideally be caught when running the relevant local tests, not when running the entire package.

On the other hand, `devtools::test()` is intended for running the full test suite once you're finished with the feature specific testing. Since we've enabled multithreaded testing, my thought was to start the slower tests first so they can run in parallel with the faster ones, theoretically, reducing the overall wall clock time.

I don't have a strong opinion though. Flexible to switch it to any other order.

OlinkAnalyze/R/check_npx.R
klevdiamanti: Since `check_npx()` is becoming optional, I'd like to leave it have the same behavior as before: return a named list with the checks it performed. There are already several users who have started using the output from `check_log` as input for the `check_log` argument of downstream functions. We should keep this backwards compatible.

OlinkAnalyze/R/olink_class.R
klevdiamanti: Done.

OlinkAnalyze/R/olink_class.R
klevdiamanti: I guess in a very edge case.

OlinkAnalyze/R/olink_class.R
klevdiamanti: This was Codex helping we out :)

OlinkAnalyze/R/olink_class.R
klevdiamanti: Done.

OlinkAnalyze/R/olink_class.R
klevdiamanti: Good point, but I think it's better to throw a warning. Implementing it.

OlinkAnalyze/R/olink_class.R
klevdiamanti: Done. Will also rename `strip_check_log()` below.

OlinkAnalyze/R/olink_class.R
klevdiamanti: Good point. Fixed.

OlinkAnalyze/R/olink_class.R
klevdiamanti: `metadata` is part of an `ArrowObject` by default:
```
x <- npx_data1
arrow::write_parquet(x, "my_file.parquet")
y <- arrow::open_dataset("my_file.parquet")
names(y)
y$metadata
# returns an empty `named list()`
```
I think we can leave it as is without any issues.

OlinkAnalyze/R/olink_class.R
klevdiamanti: Good point. Fixing it.

OlinkAnalyze/R/olink_normalization_product.R
klevdiamanti: Done.

OlinkAnalyze/R/olink_normalization_product.R
klevdiamanti: Good catch.

OlinkAnalyze/R/olink_normalization_utils.R
klevdiamanti: Fixed.

OlinkAnalyze/R/pca_plot.R
klevdiamanti: Fixed.

OlinkAnalyze/R/utils.R
klevdiamanti: We'll need to make a function that casts between ArrowObject with metadata attached and olink_class. I have it in my list, but probably we do not need to do it right now.

OlinkAnalyze/tests/testthat/helper-check.R
klevdiamanti: Position and scales were tested and added. Work with no issues. Facets, on the other hand, were failing tests because they inherit the data which in one case is olink_class and in the other case is a tbl_df. However, I fixed this by adding an extra argument `ignore_attr = TRUE`. So, all works now.

OlinkAnalyze/tests/testthat/test-check_columns.R
klevdiamanti: Done.

OlinkAnalyze/tests/testthat/test-check_npx.R
klevdiamanti: The implementation of `get_check_log()` is based in the following logic:
1. Attached `check_log` to the object.
2. Provided `check_log` by user.
3. Generate `check_log` by running `check_npx()` internally.
TBH, I do not trust the user to provide the correct `check_log`, so my take is to trust the one attached to the object given that we update it at all instances we export a dataset: `read_npx()`, `clean_npx()`, `olink_lod()` and `olink_normalization()`. Do you think we can work on an exported function `update_check_log()` or similar so that user can update their `check_log` in case they modified the NPX file themselves outside our functions?

OlinkAnalyze/tests/testthat/test-olink_anova.R
klevdiamanti: That will be done when we optimize the code for the stats functions.

OlinkAnalyze/tests/testthat/test-olink_boxplot.R
klevdiamanti: This was intentional. These were tested in the earlier `olink_boxplot - works - vdiffr` test, and as long as some basic outputs between arrow, olink_class and tbl_df agree, I think we can be confident that things work.

OlinkAnalyze/tests/testthat/test-olink_lmer.R
klevdiamanti: It works, but arrow data is converted to `tbl_df` under the hood. We will have to re-work this together with the other stats functions when we optimize them.

OlinkAnalyze/tests/testthat/test-olink_ttest.R
klevdiamanti: Correct.

OlinkAnalyze/tests/testthat/test-plot_heatmap.R
klevdiamanti: correct. this is why I added new snapshots, which are practically identical to the ones generated by `tbl_df`

OlinkAnalyze/tests/testthat/test-read_npx_zip.R
klevdiamanti: Fixed.

OlinkAnalyze/tests/testthat/test-utils.R
klevdiamanti: Fixed.

OlinkAnalyze/tests/testthat/test-utils.R
klevdiamanti: Fixed.

OlinkAnalyze/tests/testthat/test-utils.R
klevdiamanti: ArrowObject is not an S3, but an R6. Used `expect_r6_class()` instead.

OlinkAnalyze/tests/testthat/test-Olink_one_non_parametric.R
klevdiamanti: Fixed all issues.

OlinkAnalyze/tests/testthat/test-plot_is_bridgeable.R
klevdiamanti: Good catch.

OlinkAnalyze/tests/testthat/test-plot_is_bridgeable.R
klevdiamanti: Fixed.

OlinkAnalyze/tests/testthat/test-plot_is_bridgeable.R
klevdiamanti: We used to have a snapshot which we were checking against using `vdiffr`. However, the output of `olink_bridgeability_plot()` is rather unstable as it used `ggpubr::ggarrange()`, and these snapshots would very often fail. Instead, we are comparing to snapshots taken from the individual plots used to synthesize the combo plot. I would skip comparing here, and keep comparing to snapshots from the individual plot functions: `bridgeability_iqr_range_plt()`, `bridgeability_r2_plt()`, `bridgeability_counts_plt()`, `bridgeability_ks_plt()`.

OlinkAnalyze/tests/testthat/test-olink_class.R
klevdiamanti: Good point. Fixed.

OlinkAnalyze/tests/testthat/test-olink_class.R
klevdiamanti: Fixed.

OlinkAnalyze/tests/testthat/test-clean_npx.R
klevdiamanti: Great idea. Fixed.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: I think it will be excessive as this is already tested for `tbl_df`. Behavior should be the same as they use the same codebase.

OlinkAnalyze/tests/testthat/test-olink_class.R
klevdiamanti: We'll need to implement this separately.

OlinkAnalyze/R/olink_normalization.R
klevdiamanti: Chose to do it in the very beginning of `olink_norm_input_check()`:

`# Check data format
  df1_check_log <- get_check_npx(df = df1, check_log = df1_check_log)
  df1_out_df <- get_read_npx_output(df = df1)
  df1_preferred_names <- get_preferred_names(df = df1,
                                             check_log = df1_check_log)
  if (!is.null(df2)) {
    df2_check_log <- get_check_npx(df = df2, check_log = df2_check_log)
    df2_out_df <- get_read_npx_output(df = df2)
    df2_preferred_names <- get_preferred_names(df = df2,
                                               check_log = df2_check_log)
  }
`

I think it's essentially the same with your suggestion.

OlinkAnalyze/R/olink_normalization.R
klevdiamanti: Yes. This will require some research as we are using `ArrowObject` and `tbl_df` for `joins` and for performing window operations which are not very popular with `ArrowObject`. Have created card OA-2609 to address it.

OlinkAnalyze/R/olink_normalization.R
klevdiamanti: As answered in the previous comment, this will require some research a fiddle with the S3 and R6 objects. Ideally, we would like one codebase for both, so we need to look into this carefully so that `olink_normalization()` works with both `ArrowObject` and `olink_class` natively without intermediate conversions.

OlinkAnalyze/R/olink_normalization.R
klevdiamanti: I have addressed all comments regarding renaming columns. Now we are back to renaming all columns marked as `is_updateable` in `column_name_dict`, namely: `qc_warning`, `assay_warn`, `normalization`, and `qc_version`.
In addition, now `column_name_dict` also supports renaming `sample_type` and `assay_type` to make easier the downstream filtering of assays and samples.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: They all use the same codebase. I think it's excessive.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: They all use the same codebase. I think it's excessive.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: They all use the same codebase. I think it's excessive.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: They all use the same codebase. I think it's excessive.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: I've rolled back to renaming columns. Implementation differs to support S3 object.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: As long as we use the same codebase for `olink_class`, `ArrowObject` and `tbl_df` we can safely not explicitly check all possible combinations for all classes. I believe it's excessive for now.

OlinkAnalyze/tests/testthat/test-olink_normalization.R
klevdiamanti: As long as we use the same codebase for olink_class, ArrowObject and tbl_df we can safely not explicitly check all possible combinations for all classes. I believe it's excessive for now.

OlinkAnalyze/tests/testthat/test-olink_normalization_n.R
klevdiamanti: As mentioned above, I believe it would be excessive to repeat all tests from `olink_normalization` here. The issue we have here is that we end up often in situations with multiple sets of duplicated samples when we use `olink_normalization_n()`, which become problematic in handling. This is not the case with the other functions `olink_normalization_subset()` and `olink_normalization_bridge()`. We will need to discuss this a bit, but it can be done independently from this PR. Made a card about it OA-2612.

klevdiamanti:
> how is the workflow from the user's side expected to work? if a check_log is being generated on read_npx and then a user runs check_npx and provides the check_log to the function with the olink_class data then the function will use the attached check log over the user provided check log. However the workflow changes we will need to update in the overview vignette. Specifically for the case of using quant data instead of NPX in downstream analysis (as this question has already come up a few times).

Will answer this with PR updates.

klevdiamanti:
> should npx_data1 and npx_data2 be converted to olink_class instead?

Will answer this with PR updates.

klevdiamanti:
> A few changes and questions. One thing to confirm is that we aren't testing the arrow objects on the stats functions. Is this because those functions havent been updated yet.
> 
> Biggest requested change is preferred_names function which does not have the same output as the rename_cols (specifically with QC_Warning and SampleQC differing between dfs which now results in 2 columns instead of 1 combined column)

Will answer this with PR updates.

klevdiamanti:
**Answers to your previous comments on the PR:**

> should npx_data1 and npx_data2 be converted to olink_class instead?

I would not do that. I'd keep them as are. Any particular reason we would want to do this? The main purpose this upgrade serves is to make carrying check_log safer (user doesn't need to provide it every time) and user experience easier (they don't have to re-run check_npx every time).

> how is the workflow from the user's side expected to work? if a check_log is being generated on read_npx and then a user runs check_npx and provides the check_log to the function with the olink_class data then the function will use the attached check log over the user provided check log. However the workflow changes we will need to update in the overview vignette. Specifically for the case of using quant data instead of NPX in downstream analysis (as this question has already come up a few times).

The main update is that the user does not need to generate the check_log every time they update their dataset, but instead we do it for them on all the functions that edit the df. This makes it safer for OA as we do not have to rely on the user providing the correct check_log, and for the user who does not need to generate it every time.

> A few changes and questions. One thing to confirm is that we aren't testing the arrow objects on the stats functions. Is this because those functions havent been updated yet.

Yes, stats function do not natively support `ArrowObject` for the time being. We will work on this in the next sprint.

> Biggest requested change is preferred_names function which does not have the same output as the rename_cols (specifically with QC_Warning and SampleQC differing between dfs which now results in 2 columns instead of 1 combined column)

I have implemented a solution that updates the column names of the non-reference dataset to the ones of the reference dataset, for designated columns: marked as `is_updateable` in `column_name_dict`. I also expanded to rename also `sample_type` and `assay_type`, that will make downstream dataset cleanup easier.

**Additional updates on code**

Some updates I had to do that were not part of the previous version of the PR:
1. Had to request testthat v3.3.0 which is the earliest version of testthat to contain the function `expect_r6_class()` that we use to test `ArrowObject`. Made changes in `DESCRIPTION` file.
2. Because of 1 above, I had to force-install testthat v3.3.0 in earlier versions of R (4.1.3, 4.2.3, 4.3.3 and 4.4.3). Made changes in `dockerfile`.
3. `lintr` v3.4.0 caught some minor issues with single and double & in scripts `olink_anova.R`, `olink_lmer.R`, `Olink_one_non_parametric.R‎` and `olink_ordinal_regression.R‎`. These file have some minor updates to fix linting.

All other updates address comments from the review.

**Note**

Workflows `R-CI-tests-pass` and `coverage-render-and-document` still fail until PRs #773 and #774 are merged.

**Next steps**

A few upcoming updates (PRs) linked to this PR:
1. Cast from olink_class to ArrowObject: function that converts between ArrowObject and olink_class, while carrying over the check_log. OA-2614
2. Update OA main vignette. OA-2610
3. Update OAV vignettes. OA-2611
4. Function to update check log when user has updated they df themselves. OA-2613
5. Functions use native ArrowObject and do not need to fall bck to tibble: Will be investigated and fixed in upcoming releases.

PR is ready for 2nd round of review.

OlinkAnalyze/R/check_npx.R
```diff
@@ -180,169 +180,135 @@ check_npx <- function(df,
...
 #'
 #' @details
-#' This function acts as a wrapper for [`check_npx()`]. It will check if the
-#' input `check_log` provided by the user is valid. If not, it will throw
-#' relevant errors or warnings. Alternatively, if `check_log` was not provided
-#' by the user, it will run [`check_npx()`] to provide `check_log` to enable
-#' downstream functions to run.#'
+#' This function checks if the input data frame is an `olink_class` or an
+#' ArrowObject with `olink_check_log` in its metadata. If so, it uses the
+#' internal function [`olink_extract_check_log()`] to extract he `check_log`
```
kathy-nevola:
```suggestion
#' internal function [`olink_extract_check_log()`] to extract the `check_log`
```

OlinkAnalyze/data-raw/column_name_dict.R
```diff
@@ -238,8 +238,8 @@ column_name_dict <- dplyr::tibble(
   ),
   is_updatable = c(
     FALSE, # sample_id
-    FALSE, # sample_type
-    FALSE, # assay_type
+    TRUE, # sample_type
```
kathy-nevola, line 241: It might also make sense to make LOD updatable since we have had LOD, PlateLOD, and MaxLOD. We can also wait until next sprint to change this if needed for the LOD changes.

OlinkAnalyze/tests/testthat/test-clean_npx.R
```diff
@@ -2361,11 +2660,557 @@ test_that(
...
+
+    expect_identical(
+      object = olink_extract_check_log(df = curr_result_v2)$col_names,
+      expected = check_log_tmp$col_names
+    )
+  }
+)
+
+test_that(
+  "clean_npx - works - obj - if both check_log and olink_lass are provided",
```
kathy-nevola:
```suggestion
  "clean_npx - works - obj - if both check_log and olink_class are provided",
```

OlinkAnalyze/R/olink_normalization.R
```diff
@@ -381,82 +412,199 @@ olink_normalization <- function(df1,
...
-                                      not_ref_df) {
+norm_internal_update_pref_name <- function(preferred_names) {
 
-  # only these columns can be updated to the reference df
-  cols_to_update <- column_name_dict |>
+  if (is.null(preferred_names) || length(preferred_names) == 0L) {
+    return(NULL)
+  }
+
+  # `is_updatable` marks column keys whose names can be harmonized between
```
kathy-nevola, line 538: If different names are used for a non-updatable column (for example panel vs Panel), we will get the 2 columns with NA for 1 project in each column. I think this is ok for now, although maybe we can work towards more columns being updatable or at least having a warning that the column is present in one dataset and not the other.

kathy-nevola (commented):
A few minor changes but otherwise looks good.

OlinkAnalyze/data-raw/column_name_dict.R
klevdiamanti: Totally agreed for LOD, however, LOD is the only that allows multiple matches. I will skip this for now and we should make a point the card regarding LOB to address renaming of LOD.

OlinkAnalyze/R/olink_normalization.R
klevdiamanti: Totally agree. Made a card `OA-2617` to investigate and address this. I added the card to the current sprint, but if we do not manage, we can move it to the next one.

klevdiamanti:
Addressed all comments. Waiting for PR #774 to be merged, for all tests to pass and them we merge.

klevdiamanti:
Merging as PR has been practically approved by @kathy-nevola and all CI workflows are passing.

PR #771, opened 2026-06-25
Update JSON handling with bash `jq` instead of external repo dependency

klevdiamanti:
This pull request updates the workflow matrix configuration for R CMD checks, replacing the use of an external GitHub Action with an inline Bash script and updating the matrix configuration file to support multiple event types per job. The main goal is to improve maintainability and flexibility by allowing jobs to be triggered by multiple GitHub events.

**Workflow matrix generation improvements:**

* Replaced the `JoshuaTheMiller/conditional-build-matrix` GitHub Action with an inline Bash script that uses `jq` to dynamically filter the matrix based on the triggering event, increasing transparency and flexibility in matrix generation. (.github/workflows/R-CMD-check.yaml)

**Matrix configuration enhancements:**

* Updated `.github/workflows/matrix_config_check.json` to use an `events` array instead of a single `runOn` field, allowing each job to specify multiple triggering events (e.g., both `push` and `pull_request`). (.github/workflows/matrix_config_check.json)


kristianHoden (approved):
Looking good!

PR #777, opened 2026-07-20
Enable test coverage on push events

klevdiamanti:
This pull request updates the workflow conditions in `.github/workflows/coverage-render-and-document.yaml` to control when the test coverage badge and README updates occur. The main change ensures that these steps only run on pushes to specific branches, and save a lot of time as workflow runs faster in any non-develop push.

Workflow condition updates:

* Restricted the "Test coverage badge" and "Commit changes if README is updated" steps to only run on pushes to the `develop` branch or the `OA-2615-Update-README-badge-only-when-pushing-to-develop` branch, preventing unnecessary updates on other branches.


.github/workflows/coverage-render-and-document.yaml
```diff
@@ -32,12 +32,14 @@ jobs:
         with:
           extra-packages: any::pak, any::remotes, any::stringr, any::covr, any::devtools, any::rmarkdown, any::roxygen2, local::.
           working-directory: OlinkAnalyze
-
+          
       - name: Test coverage badge
+        if: github.event_name == 'push' && github.ref == 'refs/heads/develop'
```
kathy-nevola, line 37: One thought is should this maybe run on PR to main instead of push to develop? I guess it depends on how we are using the test coverage badge. Is it:
As a way for users to see how much test coverage we have and increase their confidence in our package? -> then I would suggest it is updated at new releases (PR to main or alternatively push to main)

As a way for us to track our test coverage across the package and identify areas for improvement? -> then develop push or PR makes sense since those tests have been developed, they just arent out yet.

klevdiamanti: The reason I picked `develop` is that the workflow `coverage-render-and-document` does not run on `main`.

To answer your question, the badge is mainly to provide confidence to users. Gaps in tests cannot be driven by this summary number.

kathy-nevola: makes sense to me.

kathy-nevola (approved):
LGTM

PR #782, opened 2026-07-26
Improve error message clarity for cells A1 and B1

klevdiamanti:
This pull request updates the error messaging for wide format file validation in the `read_npx_format_colnames` function and its associated tests. The new messages provide clearer guidance on the expected file structure.

**Error message improvements:**

* Updated the error message in `read_npx_format_colnames` (in `OlinkAnalyze/R/read_npx_utils.R`) to specify that only cells A1 and B1 should be populated with "Project Name" and the current project name, respectively.

**Test updates:**

* Updated the corresponding tests in `test-read_npx_utils.R` to check for the new, more descriptive error message.


kathy-nevola (approved):
LGTM

PR #783, opened 2026-07-27
Replace legacy `.Dim` with `dim`

klevdiamanti:
This pull request makes a minor change to the `test-olink_color.R` test file, updating the way the expected result's dimensions are specified to use the `dim` argument instead of `.Dim`. The legacy argument was causing a `NOTE` on CRAN check for OlinkAnalyze.

**Note:** This code should be tested using `rhub`.


kathy-nevola (approved):
LGTM. Package is now passing on win-devel CRAN server. (1 warning message from not incrementing version number but that will get fixed later on prior to release.)

PR #785, opened 2026-07-29
Add `update_check_log` function and improve documentation

klevdiamanti:
This pull request introduces a new utility to update or attach the check log metadata to Olink datasets, along with comprehensive documentation and tests. The main addition is the `update_check_log` function, which allows users to refresh or set the check log after data cleaning or column renaming, ensuring the metadata accurately reflects the current state of the dataset. The changes also include robust test coverage and user documentation.

**New feature: Check log update utility**
- Added the `update_check_log` function to refresh or attach the check log to an `olink_class` tibble or ArrowObject, supporting updates after data modifications and preferred column name changes.
- Exported the `update_check_log` function in the package namespace.

**Documentation:**
- Added detailed Roxygen and `.Rd` documentation for `update_check_log`, including usage, parameters, return value, and multiple real-world examples.

**Testing:**
- Added extensive tests for `update_check_log` covering typical use cases (refreshing logs, updating preferred names, error handling, and attaching logs to plain tibbles or ArrowObjects), ensuring correct integration and expected behavior.


OlinkAnalyze/tests/testthat/test-olink_class.R
```diff
@@ -589,6 +589,254 @@ test_that(
...
+      UniProt = LETTERS[1L:4L],
+      Assay = LETTERS[1L:4L],
+      Panel = LETTERS[1L:4L],
+      Panel_Lot_Nr = LETTERS[1L:4L],
+      NPX = rnorm(4L),
+      PlateID = rep("plate1", 4L),
+      QC_Warning = rep("Pass", 4L)
+    )
+
+    check_log <- check_npx(df = df)
```
dtopouza, line 859: With the new attach_check_log and update_check_log functions, should these be used instead of check_npx when loading a new dataset? Would we want to replace check_npx with attach_check_log entirely?

We may also want to update our vignettes with new instructions.

OlinkAnalyze/tests/testthat/test-olink_class.R
```diff
@@ -589,6 +589,254 @@ test_that(
...
+    df_arrow <- arrow::as_arrow_table(df)
+
+    expect_warning(
+      object = df_arrow_obj <- update_check_log(
+        df = df_arrow,
+        check_log = check_log
+      ),
+      regexp = "No `olink_check_log` metadata found in the ArrowObject"
```
dtopouza: This message is a bit misleading in this scenario - update_check_log adds the check_log to the arrow object, but the warning message makes it sound like the user still needs to do that ("Warning message:
No `olink_check_log` metadata found in the ArrowObject.
ℹ Consider generating it using `check_npx()`.").

Could we suppress this message from olink_extract_check_log in this particular case?

OlinkAnalyze/R/olink_class.R
```diff
@@ -424,6 +424,166 @@ rm_check_log <- function(df) {
...
+    df = df,
+    check_log = check_log
+  )
+
+  # append new preferred_names, if provided, to the existing preferred_names
+  # and remove duplicates, keeping the last occurrence of each name
+  # essentially preferred_names provided to update_check_log() will override any
+  # existing preferred_names in the check_log.
+  if (!is.null(preferred_names)) {
+    existing_preferred_names <- c(existing_preferred_names, preferred_names)
```
dtopouza, line 575: Do we want to check here that we have exactly 1 new preferred name per category? Currently it is possible to give update_check_log multiple preferred names and it will add the last one to the check log:

```
df_obj <- attach_check_log(
  df = df,
  out_df = "tibble",
  preferred_names = c("sample_id" = "SampleID2")
)

df_obj_updated <- update_check_log(
  df = df_obj,
  preferred_names = c("sample_id" = "SampleID",
                      "sample_id" = "SampleID3")
  )

olink_extract_check_log(df_obj_updated)$col_names$sample_id
# [1] "SampleID3"
```

dtopouza (approved):
Looks good overall. Added a few more unit tests and some questions for the code, but otherwise no changes needed.

OlinkAnalyze/tests/testthat/test-olink_class.R
klevdiamanti: Good catch. Fixed!

OlinkAnalyze/R/olink_class.R
klevdiamanti: Good point, I agree. `preferred_names` should not accept duplicated keys. I split the validation logic out of `check_npx_update_col_names()` into a shared helper and now call it from `update_check_log()` before merging preferred names, so duplicated entries error instead of silently keeping the last one. I also added tests for this case.

OlinkAnalyze/tests/testthat/test-olink_class.R
klevdiamanti: I would be supportive of keeping `check_npx` as some users might have implemented it in their pipelines already. Also, we use it extensively for testing purposes. I do see the use of `check_npx` declining dramatically because of the new functionality, and we can choose to make it internal-nly in the future. But for now, might be better to keep it public.

